PR #15, opened 2016-08-26
Widen scope of exceptions caught to prevent breaking out of loop and …

CarlHembrough:
### What

Improved exception handling around loading collections during pre-publish


ian-kent:
:+1:

PR #21, opened 2016-09-07
Hotfix/splunk tablebuilder issue

dave-Llew:
### What

Hot fix for spunk table builder issue. Accessing request.parameter(X) post handle caused request body input stream to be consumed too early. Moved logic to post handle.


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/filter/RequestMetricsFilter.java
```diff
@@ -18,7 +20,12 @@
     @Override
     public boolean filter(HttpServletRequest request, HttpServletResponse response) {
         if (StringUtils.isEmpty(request.getHeader(IGNORE_METRICS_HEADER))) {
-            metricsService.captureRequest(request);
+            try {
+                metricsService.captureRequest(request);
+            } catch (Exception ex) {
+                logError(ex).log();
+                return false;
```
CarlHembrough: Do we want to still return true here? We don't want any splunk issues to prevent requests being processed.

PR #25, opened 2016-09-19
Feature/approval status

CarlHembrough:
### What

Added ApprovalStatus value to collection data.

### How to review

Look at the collection json when an approval is in progress and it should be set to IN_PROGRESS. When the approval is complete the status should be COMPLETE.


Crispioso:
Checked out branch, approved collection with bulletin and showed 'in progress' status

PR #40, opened 2016-12-14
Synonyms and stemmer changes

guido4f:
Synonyms and stemmer changes
Altered Synonym's configuration to expand synonyms only at index time; disabling query time expansion.
Changed stemmer to use newer snowball stem instead of less accruate porter stemmer.

Limited to two files: 
index-config.yml
+ changed porter_stem filter to snowball filter
+ added new analyzer to be used when querying (to exclude synonym expansion)
+ set replicas to 0 as only one node in each production cluster
+ set number of shards to 3 (default is 5) as we only have one node in production,
       but 3 shards would give the opportunity to spread index over a max 3 nodes 

default-mapping.json
+ update mappings where fields use the synonym to use a new search_analyzer that does not have an Analyzer.

### DIFF
--- a/zebedee-reader/src/main/resources/search/boost.txt
+++ b/zebedee-reader/src/main/resources/search/boost.txt
@@ -12,6 +12,11 @@
 # /economy/grossdomesticproductgdp/bulletins/grossdomesticproductpreliminaryestimate/2015-10-27 => preliminary, estimates
 
 
+########### NOTES ###########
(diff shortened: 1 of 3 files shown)


ian-kent:
looks ok to me, not that most of it means much 😄 

is this something we want to go into production, or just for testing? if it's just for testing we shouldn't merge into develop branch otherwise it'll go live at some point

robchamberspfc:
Yeah these are good to go onto develop, and then live. Have checked locally but will want to have another check on develop.

PR #75, opened 2017-05-12
Feature/refactor downloads

daiLlew:
Updated generated download files are now held in memory in a byte array output stream until they are written to the http response instead of being written to disk in 'temp' files. 
In the previous implementation these temp files were never removed and there was an unnecessary number of IO read writes.
This change also removes large chunks of redundant code from data generator class, improves test coverage and other updates to follow standard java best practices.


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/DataGenerator.java
```diff
@@ -244,11 +244,11 @@ public Resource generateData(List<TimeSeries> timeSerieses, String format) throw
     byte[] csvToBytes(List<List<String>> grid) throws IOException {
         try (
                 ByteArrayOutputStream baos = new ByteArrayOutputStream();
-                CSVWriter writer = new CSVWriter(new OutputStreamWriter(baos, Charset.forName(UTF8)), ',')
+                OutputStreamWriter outputStreamWriter = new OutputStreamWriter(baos, Charset.forName(UTF8));
+                CSVWriter writer = csvWriterFactory.getCSVWriter(outputStreamWriter, ',');
```
Matt-Guest: The ByteArrayOutputStream and OutputStreamWriter could be replaced with
```
Writer writer = new CharArrayWriter()
```

daiLlew: As discussed agreed to leave as is.

PR #79, opened 2017-05-23
default to index.html, and support spaces in URIs for data vis

ian-kent:
### What

Updates to data visualisation URI handling:

* Default to `index.html` if the URI doesn't include a filename
* Support spaces in directory/file names

### How to review

* Start babbage/zebedee
* Create a visualisation (`content/visualisations/test/content/index.html`)
* Check http://localhost:8080/visualisations/test/ works
* Create a visualisation with spaces (`content/visualisations/has spaces/content/index.html`)
* Check http://localhost:8080/visualisations/has%20spaces/ works

### Who can review

Anyone except @ian-kent

### DIFF
--- a/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/CollectionReader.java
+++ b/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/CollectionReader.java
@@ -2,6 +2,7 @@ package com.github.onsdigital.zebedee.reader;
 
 import com.github.onsdigital.zebedee.content.dynamic.browse.ContentNode;
 import com.github.onsdigital.zebedee.content.page.base.Page;
+import com.github.onsdigital.zebedee.exceptions.BadRequestException;
 import com.github.onsdigital.zebedee.exceptions.NotFoundException;
 import com.github.onsdigital.zebedee.exceptions.ZebedeeException;
 import com.github.onsdigital.zebedee.reader.data.language.ContentLanguage;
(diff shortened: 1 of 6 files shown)


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/CollectionReader.java
```diff
@@ -59,7 +60,11 @@ public Page getContent(String path) throws ZebedeeException, IOException {
 
 
     public Resource getResource(String path) throws ZebedeeException, IOException {
-        return findResource(path);
+        try {
+            return findResource(path);
+        } catch (BadRequestException e) {
+            return findResource(path + "/index.html");
```
daiLlew: Do we want to check if the uri starts with /visualisations? to limit what this will be applied to? Also what happens if there is no index.html? do we get a 404?

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/ZebedeeReader.java
```diff
@@ -120,7 +121,11 @@ public Content getCollectionContent(String collectionId, String sessionId, Strin
      * @throws IOException
      */
     public Resource getPublishedResource(String path) throws ZebedeeException, IOException {
-        return publishedContentReader.getResource(path);
+        try {
+            return publishedContentReader.getResource(path);
+        } catch (BadRequestException e) {
+            return publishedContentReader.getResource(path + "/index.html");
```
daiLlew: Same here to we want to only apply this to uris starting /visualisation? I know that index.html is a convention but its not one this code follows. I am wary of applying this to everything.

ian-kent:
both index.html bits now limit it to visualisations directory

if index.html doesn't exist, the same exception is thrown which results in a 404 as before

PR #91, opened 2017-07-04
Wrapper around Serialiser to silently handle & log execeptions.

daiLlew:
Added wrapper around deserialiser class to give the option of suppressing & logging errors & if required.

Previously calling list users/sessions/teams would make the entire application fall over if one of the users/teams/sessions could not be deserialised.

This change means these errors are now caught and logged but not propagated allowing the application to carry on -> if one user file is corrupt the other users are no longer affected.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/CustomListCollector.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/CustomListCollector.java
@@ -25,7 +25,11 @@ public abstract class CustomListCollector<T, R> implements Collector<T, Immutabl
 
     @Override
     public BiConsumer<ImmutableList.Builder<T>, T> accumulator() {
-        return (builder, item) -> builder.add(item);
+        return (builder, item) -> {
+            if (item != null) {
+                builder.add(item);
+            }
+        };
     }
 
     @Override
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
@@ -1,0 +1,61 @@
+package com.github.onsdigital.zebedee.util.serialiser;
+
+import com.github.davidcarboni.restolino.json.Serialiser;
+import com.github.onsdigital.zebedee.user.model.User;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static com.github.onsdigital.zebedee.logging.ZebedeeLogBuilder.logError;
+
+/**
+ * Created by dave on 04/07/2017.
+ */
+public class JSONSerialiser<T> {
+
+    protected static final String DESERIALISE_ERROR_DETAILS_MSG = "Warning Corrupt JSON file encountered. JSON could " +
+            "not be deserialised. It is highly recommended that you investigate this cause of thus issue";
+
+    protected static final String DESERIALISE_ERROR_DETAILS_KEY = "details";
+    protected static final String DESERIALISATIOBN_ERROR_MSG = "Failed to deserialise JSON";
+
+    protected Class<T> t;
+
+    public JSONSerialiser(Class<T> t) {
+        this.t = t;
+    }
+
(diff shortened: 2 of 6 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Created by dave on 04/07/2017.
+ */
+public class JSONSerialiser<T> {
+
+    protected static final String DESERIALISE_ERROR_DETAILS_MSG = "Warning Corrupt JSON file encountered. JSON could " +
+            "not be deserialised. It is highly recommended that you investigate this cause of thus issue";
+
+    protected static final String DESERIALISE_ERROR_DETAILS_KEY = "details";
+    protected static final String DESERIALISATIOBN_ERROR_MSG = "Failed to deserialise JSON";
```
ian-kent: think  `OBN` at the end is a typo?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
```diff
@@ -0,0 +1,59 @@
+package com.github.onsdigital.zebedee.util.serialiser;
+
+import com.github.davidcarboni.restolino.json.Serialiser;
```
CarlHembrough: Is there a need to use the restolino serialiser? Consider using another.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/CustomListCollector.java
```diff
@@ -25,7 +25,11 @@
 
     @Override
     public BiConsumer<ImmutableList.Builder<T>, T> accumulator() {
-        return (builder, item) -> builder.add(item);
+        return (builder, item) -> {
+            if (item != null) {
+                builder.add(item);
+            }
+        };
```
CarlHembrough: Cannot see where this is being used. Just double check its required and does not break anything unexpected.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
```diff
@@ -0,0 +1,59 @@
...
+        } catch (Exception e) {
+            logError(e, DESERIALISATION_ERROR_MSG)
+                    .addParameter(DESERIALISE_ERROR_DETAILS_KEY, DESERIALISE_ERROR_DETAILS_MSG)
+                    .path(p.toString())
+                    .log();
+            return null;
+        }
+    }
+
+    public T deserialiseQuietly(InputStream inputStream, Path p) {
```
CarlHembrough: Can we remove this overload and just have the function that takes a path ? It looks like the only places that use this function open the file just before the call. It will be cleaner if you can remove this method, then remove the opening of files around the calls to it and just pass the path in?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/CustomListCollector.java
daiLlew: This class allows you to use streams with UserList and TeamsList so its not used in changes iv made here but in the code that uses the collector - I did this change a while ago but I added the null check because it fell over when I made this change.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
daiLlew: I could be given that everywhere else uses it it wouldn't follow the current "style" and also could potentially cause some weirdness if does stuff differently i.e. serialise with one deserialise wit the other.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/serialiser/JSONSerialiser.java
daiLlew: I think I added it so the calling code didn't have to change in a few places didn't have to change.

PR #99, opened 2017-08-14
Feature/data viz preview

daiLlew:
### What

Stripped out Data Vis user type and implemented changes to allows data vis preview.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Content.java
```diff
@@ -9,30 +9,30 @@
...
 
 public class Content {
 
+
     public static final String REDIRECT = "redirect.txt";
     public static final String DATA_VIS_DIR = "visualisations";
     public static final String TIME_SERIES_KEYWORD = "timeseries";
 
+    private static final Predicate<Path> IS_DATA_VIZ_FILE = (p) -> p != null && p.toFile().isDirectory() &&
+            DATA_VIS_DIR.equals(p.getFileName().toString());
```
CarlHembrough: The original function to determine if its a data viz directory also checked the parent directory. Does this not need to be done? 

```
private static boolean isDataVisualisation(Path p) {
        // should be under visualisations but stop at the content directory.
        return p.getFileName().toString().equals(DATA_VIS_DIR) || p.getParent().getFileName().toString().equals(DATA_VIS_DIR);
    }
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Content.java
```diff
@@ -417,19 +410,19 @@ boolean moveFile(String fromUri, String toUri) throws IOException {
...
-            // PUBLISHING SUPPORT
-            return Files.isDirectory(entry)
-                    && !isTimeseries(entry)
-                    && isNotPreviousVersions(entry)
-                    && !isDataVisualisation(entry);
+    static boolean isVisible(Path entry) {
+        return Files.isDirectory(entry)
+                && !isTimeseries(entry)
+                && isNotPreviousVersions(entry)
+                && !isDataVisSubDir(entry);
```
CarlHembrough: Don't we want to show the data vis directories now? this is preventing them from being shown?

daiLlew: In the browse tree we only want to how the top level of the zip. e.g. /visualisations/dvc123 we dont want all the sub dirs and files to display

zebedee-reader/src/main/resources/logback.xml
```diff
@@ -41,6 +41,10 @@
         <appender-ref ref="PLAIN_TEXT"/>
     </logger>
 
+    <logger name="com.github.davidcarboni.restolino" level="warn" additivity="false">
+        <appender-ref ref="PLAIN_TEXT"/>
+    </logger>
+
```
CarlHembrough: Do we want this on production environment or was it only for local dev / debug?

PR #109, opened 2017-10-20
Feature/dataset metadata

CarlHembrough:
### What

* When a dataset is added to a collection, call the Dataset API and set the collection ID and state fields on the dataset. (state set to 'associated')
* When a dataset is removed from a collection, call the Dataset API to reset the collection ID and state fields. (state set back to 'created' from 'associated')
* When a dataset version is added to a collection, call the Dataset API and set the collection ID and state fields on the dataset version. (state set to 'associated')
* When a dataset version is removed from a collection, call the Dataset API to reset the collection IF and state fields. (state set back to 'created' from 'associated')

### How to review

Review code

### Who can review

Anyone familiar with Java.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/dataset/api/DatasetAPIClient.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/dataset/api/DatasetAPIClient.java
@@ -3,10 +3,13 @@ package com.github.onsdigital.zebedee.dataset.api;
 import com.github.onsdigital.zebedee.configuration.Configuration;
 import com.github.onsdigital.zebedee.content.util.ContentUtil;
 import com.github.onsdigital.zebedee.dataset.api.exception.BadRequestException;
+import com.github.onsdigital.zebedee.dataset.api.exception.DatasetAPIException;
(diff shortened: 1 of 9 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/dataset/api/DatasetAPIClient.java
```diff
@@ -98,13 +100,56 @@ public Dataset getDataset(String datasetID) throws IOException, DatasetNotFoundE
...
+     * @throws DatasetAPIException
+     */
+    @Override
+    public DatasetVersion getDatasetVersion(String datasetID, String edition, String version) throws IOException, DatasetAPIException {
+
+        if (StringUtils.isEmpty(datasetID) || StringUtils.isEmpty(edition)) {
+            throw new BadRequestException("A dataset ID, edition, and version must be provided.");
+        }
+
+        String path = "/datasets/" + datasetID + "/editions/" + edition + "/versions/" + version ;
```
Matt-Guest: Is it worth using a URL builder?

CarlHembrough: Good point - I have refactored to use the builder.

PR #113, opened 2017-11-14
Feature/add dataset to collection

CarlHembrough:
### What

* Improved logging of requests to the dataset API
* Improved handling of missing Dataset links (null checks)
* Populate the URL for the dataset version

### How to review

Review code changes

### Who can review

Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeDatasetService.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeDatasetService.java
@@ -15,6 +15,8 @@ import com.github.onsdigital.zebedee.util.ZebedeeCmsService;
 import java.io.IOException;
 import java.util.Optional;
 
+import static com.github.onsdigital.zebedee.logging.ZebedeeLogBuilder.logInfo;
+
 /**
  * Dataset related services
  */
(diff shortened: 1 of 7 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeDatasetService.java
```diff
@@ -51,27 +53,39 @@ public CollectionDataset updateDatasetInCollection(String collectionID, String d
...
+
+            collectionDataset.setTitle(dataset.getTitle());
+
+            if (dataset.getLinks() != null && dataset.getLinks().getSelf() != null) {
+                collectionDataset.setUri(dataset.getLinks().getSelf().getHref());
+            } else {
+                logInfo("The dataset URL has not been set on the dataset response.")
+                        .addParameter("collectionID", collectionID)
+                        .addParameter("datasetID", datasetID)
+                        .log();
```
Matt-Guest: Should this throw an exception? So The HTTP handler can send the correct code?

CarlHembrough: Yes you are right. I had been ignoring this case while a bug gets fixed on the dataset API, but its correct to just throw the exception if its not there. 👍

PR #119, opened 2018-01-08
make auditdb configurable from environment variables

ian-kent:
### What

- allow auditdb to be configured from environment variables
- removes unused config option from run.sh

### How to review

- start zebedee, get postgres connection errors
- set `db_audit_url`, `db_audit_username` and `db_audit_password`
- start zebedee, don't get postgres connection errors

### Who can review

Anyone except @ian-kent


LloydGriffiths:
👍

PR #120, opened 2018-01-11
Add teams with IDs to collectionDetails endpoint

Crispioso:
### What

Includes two changes:
1) Add teams with IDs to collectionDetails endpoint for the refactored Florence collections screen.
2) Remove the bug where team names were duplicated in the 'teams' property on the collectionDetails response.

### How to review

1) Switch to branch `feature/add-teams-details-to-collection` on Zebedee.
2) Ensure you have a collection with at least one team added to it.
3) Ensure the added teams has at least one member.
4) Get a response from the `collectionDetails` endpoint on Zebedee for that collection (e.g. `/collectionDetails/mycollection-mycollectionsuid`).
5) Check that the `teams` property only has one entry for each added team, not two.
6) Check that the `teamsDetails` property has an array for object with the same teams but instead including the team ID and an empty members array.

### Who can review

Probably @daiLlew, but maybe @ian-kent?

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
@@ -79,17 +79,7 @@ public class CollectionDetails {
         addEventsForDetails(result.reviewed, result, collection);
 
         Set<Integer> teamIds = Root.zebedee.getPermissionsService().listViewerTeams(collection.description, session);
-        List<Team> teams = Root.zebedee.getTeamsService().resolveTeams(teamIds);
-
-        // The members property will always be empty because we're using the Team class, rather than creating a new one
-        // that doesn't include members
-        result.teamsDetails = teams.stream().map(team -> {
-            Team resultTeam = new Team();
-            resultTeam.setId(team.getId());
-            resultTeam.setName(team.getName());
-            return resultTeam;
-        }).collect(Collectors.toList());
-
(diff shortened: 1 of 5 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
```diff
@@ -79,9 +80,15 @@ public CollectionDetail get(HttpServletRequest request, HttpServletResponse resp
 
         Set<Integer> teamIds = Root.zebedee.getPermissionsService().listViewerTeams(collection.description, session);
         List<Team> teams = Root.zebedee.getTeamsService().resolveTeams(teamIds);
-        teams.forEach(team -> {
-            collection.description.teams.add(team.getName());
-        });
+
+        // The members property will always be empty because we're using the Team class, rather than creating a new one
```
daiLlew: Not sure I follow this. Doesn't teams ^^^ have populated values already? Why do we need to create a new list for the team details? Why can't we just use the teams?

Crispioso: Fair question. The main reasons were (and I'm happy to be told I'm wrong on this):
1) We're not 100% sure what in old Florence is using the existing teams array and there's a high chance we'd unknowingly break something.
2) The teams property has a type of `List<String>`, which doesn't match our use (we need `List<Team>` at least) but it's in the base collection class, which I don't particularly want to edit because it's extended by quite a few other classes.

PR #138, opened 2018-03-05
Datasets review workflow

Crispioso:
### What

Add in 'last edited by' information and enforce a workflow. The workflow is simplified, for the sake of getting CMD working end-to-end, but there will still be some rules:

- The user who submitted something for review cannot review their own content changes.
- Once a page has been reviewed it always stays in 'reviewed' even if there are edits to it.
- If a page is saved when 'in progress' or 'awaiting review' then it goes to 'in progress' state and the user who last edited it is stored in the collection JSON data.
- Every save/submit for reviewed/mark as reviewed stores who performed that action in `lastEditedBy` property in the a dataset/version object in the collection JSON.

### How to review

- Checkout `feature/datasets-review-workspace` on Zebedee and Florence.
- Make sure you have a dataset and a new version available to edit. The following steps should work identically with both.

Check that a user can't move their own content changes from the states 'Completed` or `InProgress` to `Reviewed` using the browser console:

1) In the Chrome Devtools console try to update the state of your own in progress or completed page to reviewed. For example
```
fetch("/zebedee/collections/my-collection-name-6ed44c3dbd1304ceb25eb22546c24c5007d5052c1d3cf1e9c0ec5c0682978aea/datasets/mid-year-pop-est/editions/time-series/versions/1", {
	credentials: 'include',
	method: 'PUT',
	body: JSON.stringify({
		state: "Reviewed"		
    })
}).then(response => console.log(response));
```
2) The above code should give a 403 response if you're trying to approve your own changes.
3) Log out and login as a different user. Now attempt to do the same as above in the console but with the different user. It should give a 200 response and the collection JSON data is updated with a 'Reviewed' state for that dataset/version.

### Who can review

Zebedee specifically should have someone like @daiLlew, @CarlHembrough or @ian-kent look at it.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/CollectionDataset.java
```diff
@@ -42,6 +43,14 @@ public void setUri(String uri) {
...
+        this.lastEditedBy = user;
+    }
+
+    public String getLastEditedBy() {
+        return lastEditedBy;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
```
daiLlew: I dont know what the equals method is used for in the code base but if you have added a new field to this class does that change the definition of equality? Basically do we want to factor `lastEditedBy`into deciding if 2 CollectionDatasets are equal.
If you do update the equals method you should also update hashCode to take into account the new field.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/CollectionDatasetVersion.java
```diff
@@ -60,6 +61,14 @@ public void setUri(String uri) {
...
+        this.lastEditedBy = user;
+    }
+
+    public String getLastEditedBy() {
+        return lastEditedBy;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
```
daiLlew: Same in this clas

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeDatasetService.java
```diff
@@ -28,11 +30,44 @@ public ZebedeeDatasetService(DatasetClient datasetClient) {
         this.datasetClient = datasetClient;
     }
 
+    private ContentStatus updatedStateInCollection(ContentStatus currentState, ContentStatus newState, String lastEditedBy, String user) throws ForbiddenException, BadRequestException {
```
daiLlew: I know this is MVP but I would consider checking if each of method params is null before doing anything and throwing the appropriate exception.

Or you could just update the order of values in the equals to make it null safe - just swap the order of the values so you are comparing the constant to the variable. For example
```
if (currentState == null && ContentStatus.Reviewed.equals(newState)) {...}
```
In this case even if `newState` is null it wont throw an null pointer exception.

Also I would consider adding logging to each case (error or a valid state update) - sees like something we want to trace & possible audit. I'll leave that up to you but if there is a bug logging will be super useful...

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collections.java
```diff
@@ -225,22 +227,22 @@ private void updateDatasetVersionInCollection(Collection collection, String data
...
-            datasetService.updateDatasetVersionInCollection(collection, datasetID, edition, version, datasetVersion);
+            datasetService.updateDatasetVersionInCollection(collection, datasetID, edition, version, datasetVersion, user);
 
         } catch (JsonSyntaxException ex) {
             throw new BadRequestException(ex.getMessage());
         }
     }
 
-    private void updateDatasetInCollection(Collection collection, String datasetID, HttpServletRequest request) throws ZebedeeException, IOException, DatasetAPIException {
+    private void updateDatasetInCollection(Collection collection, String datasetID, HttpServletRequest request, String user) throws ZebedeeException, IOException, DatasetAPIException {
```
daiLlew: Can you add the user to the logParameters as well please (there is already a `.user(String email)` on the log builder so no need to do `addParameter`.  Can you do this in all places where you have added the user field that is (already logging something).

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/CollectionDataset.java
Crispioso: Thanks, I hadn't considered that. I think considering we don't seem to mind about the other fields for equality (i.e. it's only dependent on the ID being the same) I assume the `lastEditedBy` field shouldn't either.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeDatasetService.java
Crispioso: We're already doing a null check before we call this function:
https://github.com/ONSdigital/zebedee/pull/138/files/5c7294f263ca380ea44b722e9078ed69b48ab322#diff-bbf23d72826f8b295a4591a7f223131cL112

However, make sure it never fails, even if it's called elsewhere, I've added a null check into the `updatedStateInCollection` block.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collections.java
Crispioso: Should be covered in commit https://github.com/ONSdigital/zebedee/pull/138/commits/bd73d27b2039351dfc29e46774658062904d9282

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeDatasetService.java
Crispioso: I've added some logging now too:
https://github.com/ONSdigital/zebedee/pull/138/commits/edb0f5306d1d58c0a7c0f7a48ce73a03e8140254

PR #140, opened 2018-03-07
initial changes to support user identity

daiLlew:
Created a new api endpoint to get user identity and permissions from their auth token.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,74 @@
...
+			response.setCharacterEncoding(CHAR_ENCODING);
+			response.getWriter().write(body.toJSON());
+		} catch (IOException e) {
+			logError(e, "error while attempting to write userIdentity to HTTP response").log();
+			throw e;
+		}
+	}
+
+	private AuthorisationService getAuthorisationService() {
+		if (authorisationService == null) {
```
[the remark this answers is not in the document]
daiLlew: I don't think so - I am not sure how Restolino instantiates the API classes but there is no state in the service anyway so if its overwritten there isn't anything to go wrong. I went with lazy init to make it easier to test

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/IdentityTest.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.apache.http.HttpStatus.SC_UNAUTHORIZED;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Mockito.when;
+
+public class IdentityTest {
+
+	private static final String JSON_CONTENT_TYPE = "application/json";
+	private static final String CHAR_ENCODING = "UTF-8";
```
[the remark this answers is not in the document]
daiLlew: Done - also updated to use `javax.ws.rs.core.MediaType.APPLICATION_JSON` just for good measure

zebedee-cms/src/main/resources/swagger/identity-api-swagger.yml
```diff
@@ -0,0 +1,45 @@
...
+definitions:
+  UserIdentity:
+    description: "The identity of the authenticated user"
+    type: object
+    properties:
+      email:
+        example: "tyroin@lannister.com"
+        type: string
+  Error:
+    description: "internal server error"
```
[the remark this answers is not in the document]
daiLlew: For this api this will only ever be a single error as its fail fast. However, if it's to consistent with the other API's I am happy to follow suit

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,74 @@
...
+import static com.github.onsdigital.zebedee.logging.ZebedeeLogBuilder.logInfo;
+import static com.github.onsdigital.zebedee.logging.ZebedeeLogBuilder.logWarn;
+import static org.apache.http.HttpStatus.SC_OK;
+import static org.apache.http.HttpStatus.SC_UNAUTHORIZED;
+
+@Api
+public class Identity {
+
+	private static final String JSON_CONTENT_TYPE = "application/json";
+	private static final String CHAR_ENCODING = "UTF-8";
```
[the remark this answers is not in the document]
daiLlew: done.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,74 @@
...
+
+		try {
+			UserIdentity identity = getAuthorisationService().identifyUser(sessionID);
+			logInfo("authenticated user identity confirmed")
+					.sessionID(sessionID)
+					.user(identity.getEmail())
+					.log();
+			writeResponse(response, identity, SC_OK);
+		} catch (UserIdentityException e) {
+			writeResponse(response, new Error(e.getMessage()), e.getResponseCode());
```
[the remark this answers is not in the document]
daiLlew: I did consider it but specific details of the error will be logging inside `authorisationService.identifyUser()` which will probably include more detail than the error response. I will add a log line to say returning error response

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,127 @@
...
+	private AuthorisationService authorisationService;
+    private ServiceStore serviceStore;
+
+    static final String AUTHORIZATION_HEADER = "Authorization";
+
+    @GET
+	public void identifyUser(HttpServletRequest request, HttpServletResponse response) throws IOException,
+			UserIdentityException {
+
+		if (request.getHeader(AUTHORIZATION_HEADER) != null) {
```
daiLlew: Replace null check with `StringUtils.isNotEmpty` to cover both null and empty cases

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,127 @@
...
+			writeResponse(response, identity, SC_OK);
+		} catch (UserIdentityException e) {
+			logError(e, "identify user failure, returning error response").log();
+			writeResponse(response, new Error(e.getMessage()), e.getResponseCode());
+		}
+	}
+
+	private ServiceAccount findService(HttpServletRequest request) throws IOException {
+        final ServiceStore serviceStore = getServiceStoreImpl();
+		String authorizationHeader = request.getHeader("Authorization");
```
daiLlew: Use constant already defined.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,127 @@
...
+		if (authorizationHeader != null) {
```
daiLlew: use `StringUtils.isNotEmpty` or equivalent to cover null & empty cases.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,127 @@
...
+			authorizationHeader = authorizationHeader.toLowerCase();
+			if (authorizationHeader.toLowerCase().startsWith("bearer ")) {
+				String[] cred = authorizationHeader.split("bearer ");
+				String token = cred[1];
```
daiLlew: might be a good idea to explicitly check the array len == 2  here and return an error response - not authenticated if not. Unlikely to happen but just happen but its cleaner and we are explicitly handling that error then.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -0,0 +1,127 @@
...
+				final ServiceAccount service = serviceStore.get(token);
+				if (service != null) {
+					logInfo("authenticated service account confirmed")
+							.user(service.getId())
+							.log();
+					return service;
+				} else {
```
daiLlew: I think the `else` can be removed here just let the null at the end of the method return.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private ServiceStore serviceStore;
+
+    private SessionsService sessionsService;
+
+    private PermissionsService permissionsService;
+
+    @POST
+    public ServiceAccountWithToken createService(HttpServletRequest request, HttpServletResponse response) throws IOException, NotFoundException, UnauthorizedException {
+        final Session session = getSessionsService().get(request);
```
daiLlew: I think this can return null - which would cause a null pointer on session.getEmail(), can you add a check for this and return the appropriate response.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
```diff
@@ -0,0 +1,85 @@
...
+        PermissionDefinition permissionDefinition = getPermissionsService().userPermissions(session.getEmail(), session);
```
daiLlew: Same here - although I am not sure if this one can return null - can you check and handle if need be.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
```diff
@@ -0,0 +1,85 @@
...
+        if (BooleanUtils.isTrue(permissionDefinition.admin)) {
```
daiLlew: There is a an `isAdministrator(Session session)` method on the permission service which would make this nicer.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
```diff
@@ -0,0 +1,85 @@
...
+            final ServiceStore serviceStoreImpl = getServiceStore();
+            final String token = randomIdGenerator.get();
+            ServiceAccount service = serviceStoreImpl.store(token, request.getInputStream());
```
daiLlew: Do you need to put the input steam in a try-with-resources? Im never sure when its a request inputstream. Also can this throw an exception? If so we should handle the error case and log/return the appropriate response.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private PermissionsService getPermissionsService() {
+        if (permissionsService == null) {
+            permissionsService = Root.zebedee.getPermissionsService();
+        }
+        return permissionsService;
+    }
+
+    private class ServiceAccountWithToken  extends ServiceAccount{
```
daiLlew: If you implement JSONable you can then pass this object into the `writeResponse` method directly - which means that both user identity and service identity are handled the same in the handler which feels a bit nicer.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
```diff
@@ -27,6 +27,8 @@
 			.add(ReIndex.class)
 			.add(Ping.class)
 			.add(ClickEventLog.class)
+			.add(Identity.class)
+			.add(Service.class)
```
daiLlew: I think these should be authenticated endpoints.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ServiceStoreImpl.java
```diff
@@ -0,0 +1,55 @@
...
+                service = jsonSerialise.deserialiseQuietly(input, path);
+            }
+        }
+        return service;
+    }
+
+    @Override
+    public ServiceAccount store(String token, InputStream service) throws IOException {
+        final Path path = rootLocation.resolve(token + JSON_EXTENSION);
+        ServiceAccount object = jsonSerialise.deserialiseQuietly(service, path);
```
daiLlew: Do you need to check if it already exists?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
Matt-Guest: The service endpoint shouldn't be there, but the identity endpoint should still work without the florence token

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
Matt-Guest: The rule i always follow is, who ever creates the resources it should be there responsibly to clean up. As restolina creates the request, i would expect it to handle closing the stream.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ServiceStoreImpl.java
Matt-Guest: Good point 👍, I added a check in

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -200,6 +195,8 @@ public Path getRedirectPath() {
         return redirectPath;
     }
 
+    public Path getServicePath() { return servicePath;}
```
mattrout92: To be consistent with other methods in this file, probably want to multi-line this

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
```diff
@@ -0,0 +1,76 @@
...
+    @POST
+    public void createService(HttpServletRequest request, HttpServletResponse response) throws IOException, NotFoundException, UnauthorizedException {
+        final Session session = getSessionsService().get(request);
+        if (session != null && getPermissionsService().isAdministrator(session)) {
+            final ServiceStore serviceStoreImpl = getServiceStore();
+            final String token = randomIdGenerator.get();
+            ServiceAccount service = serviceStoreImpl.store(token, request.getInputStream());
+            logInfo("new service account created").addParameter("id", service.getId()).log();
+            writeResponse(response, new ServiceAccountWithToken(token, service.getId()), SC_CREATED);
+
```
mattrout92: Should there be a return here to avoid the status getting set to forbidden?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/AuthorisationServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		Session session;
+		try {
+			session = sessionServiceSupplier.getService().get(sessionID);
+		} catch (IOException e) {
+			logError(e, "identify user error, unexpected error while attempting to get user session")
+					.sessionID(sessionID)
+					.log();
+			throw new UserIdentityException("internal server error", SC_INTERNAL_SERVER_ERROR);
```
mattrout92: could the "internal server error" string a constant. Same for the other error messages

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/UserIdentityException.java
```diff
@@ -0,0 +1,27 @@
...
+public class UserIdentityException extends ZebedeeException {
+
+    public UserIdentityException(String message, int responseCode) {
+        super(message, responseCode);
+    }
+
+    public int getResponseCode() {
+        return super.statusCode;
+    }
+
```
mattrout92: white space

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Service.java
Matt-Guest: Yes, thanks for that 👍

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/UserIdentityException.java
Matt-Guest: done

PR #147, opened 2018-03-19
Feature/update nomad plan

janderson2:
### What

* Set the max and min heap size to the same to avoid wasted time reallocating memory to the heap
* Enable auto rollback of failed nomad deployments
* Increase the nomad stagger time to allow longer for startup

### How to review

Ensure configurations are valid and logical.

### Who can review

Anyone but me.


janderson2:
@LloydGriffiths I did consider that, but I figured since the default `checks` uses both (if present) I did not see any gain.

PR #159, opened 2018-04-11
Feature/logging

daiLlew:
### What
Added detailed logging around the publishing process. Publish should continue to work exactly the same but the log output should now give enough information to specifically identify:

- the transaction IDs and the instance of the train they belong to.
- the collection ID
- the uri of the content

Error responses should now also log out more details - the endpoint called, the URL of the train instance as well the collection ID and transaction ID

I have also tried to simplify the Publisher class by collapsing the many small methods for each step of the process into one.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
@@ -115,7 +115,7 @@ public class Root {
         //Setting zebedee root as system property for zebedee reader module, since zebedee root is not set as environment variable on develop environment
         System.setProperty(ZEBEDEE_ROOT, root.toString());
 
-        //SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
+        SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
 
         loadExistingCollectionsIntoScheduler();
         initialiseCsdbImportKeys();
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
@@ -115,7 +115,7 @@ public class Root {
         //Setting zebedee root as system property for zebedee reader module, since zebedee root is not set as environment variable on develop environment
         System.setProperty(ZEBEDEE_ROOT, root.toString());
 
-        SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
+        //SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
 
         loadExistingCollectionsIntoScheduler();
         initialiseCsdbImportKeys();
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/configuration/Configuration.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/configuration/Configuration.java
@@ -88,7 +88,7 @@ public class Configuration {
     }
 
     public static List<Host> getWebsiteHosts() {
-        return Arrays.asList(StringUtils.split(defaultIfBlank(getValue("website_url"), DEFAULT_TRAIN_URL), ","))
+        return Arrays.asList(StringUtils.split(defaultIfBlank(getValue("website_url"), DEFAULT_WEBSITE_URL), ","))
                 .stream()
                 .map(url -> new Host(url))
                 .collect(Collectors.toList());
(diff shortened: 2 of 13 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/configuration/Configuration.java
```diff
@@ -74,12 +80,18 @@ public static String getMathjaxServiceUrl() {
...
+        return Arrays.asList(StringUtils.split(defaultIfBlank(getValue("publish_url"), DEFAULT_TRAIN_URL), ","))
+                .stream()
+                .map(url -> new Host(url))
+                .collect(Collectors.toList());
     }
 
-    public static String[] getWebsiteUrls() {
-        return StringUtils.split(StringUtils.defaultIfBlank(getValue("website_url"), DEFAULT_WEBSITE_URL), ",");
+    public static List<Host> getWebsiteHosts() {
+        return Arrays.asList(StringUtils.split(defaultIfBlank(getValue("website_url"), DEFAULT_TRAIN_URL), ","))
```
daiLlew: CORRECT DEFAULT

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/logging/ZebedeeLogBuilder.java
```diff
@@ -47,6 +55,10 @@
     private static final String BLOCKING_COLLECTION = "blockingCollection";
     private static final String TARGET_PATH = "targetPath";
     private static final String TARGET_COLLECTION = "targetCollection";
+    private static final String TRAIN_HOST = "trainHost";
+    private static final String TRAIN_HOSTS = "trainHosts";
+    private static final String TRANSACTION_ID = "transactionID";
+    private static final String TRANSACTION_IDS = "transactionIDS";
```
CarlHembrough: snake case for log parameter keys?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PostPublisher.java
```diff
@@ -364,7 +364,10 @@ private static void reIndexPublishingSearch(final String uri) throws IOException
 
     public static void copyFilesToMaster(Zebedee zebedee, Collection collection, CollectionReader collectionReader)
             throws IOException, ZebedeeException {
-        logInfo("Moving files from collection into master").collectionName(collection).log();
+        logInfo("Moving files from collection into master")
```
CarlHembrough: lower case?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PublishNotification.java
```diff
@@ -52,27 +49,39 @@ public PublishNotification(Collection collection) {
...
     public void sendNotification(EventType eventType) {
-        logInfo("Sending publish notification to website").addParameter("eventType", eventType.name()).log();
+        Host host = null;
         try (Http http = new Http()) {
-            for (Host host : websiteHosts) {
+            for (Host h : websiteHosts) {
+                host = h;
+                logInfo("sending publish notification to website host")
+                        .collectionId(payload.collectionId)
+                        .addParameter("websiteHost", host.toString())
```
CarlHembrough: Other go projects are using snake casing, but I guess if Zebedee is already using camel casing it may be worth sticking with that

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -480,38 +488,55 @@ private static void publishFile(
...
         if (response.statusLine.getStatusCode() != 200) {
             int code = response.statusLine.getStatusCode();
             String reason = response.statusLine.getReasonPhrase();
             String message = response.body != null ? response.body.message : "";
+
+            URI uri = endpoint.url();
+            logWarn("PUBLISH: request was unsuccessful")
+                    .transactionID(TransactionID)
+                    .collectionId(collectionID)
+                    .addParameter("trainHost", uri.getScheme() + "://" + uri.getHost() + ":" + uri.getPort())
```
CarlHembrough: You can pass `uri` in on its own and it will produce the same result, instead of building the string from parts

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
```diff
@@ -115,7 +115,7 @@ public static void init() {
         //Setting zebedee root as system property for zebedee reader module, since zebedee root is not set as environment variable on develop environment
         System.setProperty(ZEBEDEE_ROOT, root.toString());
 
-        //SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
+        SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
```
gedge: sad lack of gif 😢

PR #162, opened 2018-05-10
Hotfix/create trans defect

daiLlew:
Fix for create publishing transaction issue.

Shared `Http` object was being closed early when running against multiple train hosts before all requests had finished using it. Each thread now has its own instance of `http` to prevent this from happening.

@ian-kent @janderson2


daiLlew:
merged

PR #164, opened 2018-05-11
added fix for null pointer in publish notification to slack

daiLlew:
Collection publish notification to slack is throwing a null pointer exception if the publish was unsuccessful - added defensive checks to prevent this.

Also updated logic to send publish fail message if unsuccessful.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
@@ -20,6 +20,7 @@ import java.util.concurrent.Future;
 
 import static com.github.onsdigital.zebedee.logging.ZebedeeLogBuilder.logDebug;
 import static com.github.onsdigital.zebedee.logging.ZebedeeLogBuilder.logError;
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -115,49 +116,50 @@ public static void send(String message, String slackToken, String slackChannel,
...
-                " was published at " + format.format(publishedCollection.publishStartDate) +
-                " with " + result.transaction.uriInfos.size() + " files " +
-                " in " + timeTaken + " seconds. Example Uri: http://www.ons.gov.uk" + exampleUri;
+            result.transaction.uriInfos
+                    .stream()
+                    .filter(info -> info.uri.endsWith("data.json"))
+                    .findFirst()
+                    .ifPresent(urlInfo -> {
+                        msg.append("Example Uri: http://www.ons.gov.uk")
+                                .append(urlInfo.uri.substring(0, urlInfo.uri.length() - ("data" + ".json").length()));
```
janderson2: `"data" + ".json" does not seem to be adding anything here.

PR #169, opened 2018-05-29
Tell user which collection a URI is being edited in

ian-kent:
### What

Tell user which collection a URI is being edited in: 

* Add version of `checkForCollectionBlockingChange` which doesn't require a collection context
* Add `/checkcollectionsforuri` endpoint which returns a `204` if the URI is safe to edit, or a `200` and the collection name if the URI is being edited in another collection
* Store the collection name in a `DeleteContentRequestDeniedException`
* Update `ServerResponse` to allow additional data to be returned
* Add`ZebedeeExceptionWithData` which subclasses `ZebedeeException` and allows additional data (as a `Map<String,String>`)
* Allow `ConflictException` to store the collection name, and update it to use `ZebedeeExceptionWithData`
* Update collections handling to return the collection name when there's a URI conflict

### How to review

* Use the `feature/already-in-collection` branch of florence
* Edit URI in one collection and save, edit the same URI in another collection and you should get an error telling you it's in the first collection
* Create new URI in one collection, create the same URI again in another collection, you should get an error telling you it's in the first collection

### Who can review

Anyone except @ian-kent

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CheckCollectionsForURI.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CheckCollectionsForURI.java
@@ -1,0 +1,48 @@
+package com.github.onsdigital.zebedee.api;
+
+import com.github.davidcarboni.restolino.framework.Api;
+import com.github.onsdigital.zebedee.exceptions.UnauthorizedException;
+import com.github.onsdigital.zebedee.exceptions.ZebedeeException;
+import com.github.onsdigital.zebedee.session.model.Session;
+import com.github.onsdigital.zebedee.util.ZebedeeCmsService;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.GET;
+import java.io.IOException;
+import java.util.Optional;
+
+@Api
+public class CheckCollectionsForURI {
+    private static ZebedeeCmsService zebedeeCmsService = ZebedeeCmsService.getInstance();
+
+    /**
+     * Checks whether a URI is already in a collection
+     *
+     * @param request This should contain a X-Florence-Token header for the current session
+     * @param response 200 if the URI is being edited, othewise a 204
+     * @return The collection name if the URI is being edited, otherwise an empty string
+     * @throws IOException If an error occurs in processing data, typically to the filesystem, but also on the HTTP connection.
+     * @throws UnauthorizedException If the user does not have editor/admin permission.
+     */
+    @GET
+    public String get(HttpServletRequest request, HttpServletResponse response)
+            throws IOException, UnauthorizedException {
+
+        Session session = Root.zebedee.getSessionsService().get(request);
+        if (session == null || !zebedeeCmsService.getPermissions().canEdit(session.getEmail())) {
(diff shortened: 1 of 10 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CheckCollectionsForURI.java
```diff
@@ -0,0 +1,48 @@
...
+    @GET
+    public String get(HttpServletRequest request, HttpServletResponse response)
+            throws IOException, UnauthorizedException {
+
+        Session session = Root.zebedee.getSessionsService().get(request);
+        if (session == null || !zebedeeCmsService.getPermissions().canEdit(session.getEmail())) {
+            throw new UnauthorizedException("You are not authorised to check collections for a URI");
+        }
+
+        String uri = request.getParameter("uri");
```
daiLlew: This might be handled by the code it calls into but what happens if uri is empty?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -577,6 +577,11 @@ public void writeContent(
             // create the file
             if (!collection.create(session.getEmail(), uri)) {
                 // file may be being edited in a different collection
+                Optional<Collection> otherCollection = zebedeeSupplier.get().checkForCollectionBlockingChange(uri);
+                if(otherCollection != null && otherCollection.isPresent()) {
```
daiLlew: do you need the `!= null` I think optional should always return an value so you should be able to just do the `isPresent` check.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -585,6 +590,11 @@ public void writeContent(
             boolean result = collection.edit(session.getEmail(), uri, collectionWriter, recursive);
             if (!result) {
                 // file may be being edited in a different collection
+                Optional<Collection> otherCollection = zebedeeSupplier.get().checkForCollectionBlockingChange(uri);
+                if(otherCollection != null && otherCollection.isPresent()) {
```
daiLlew: same here.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
ian-kent: I thought that, but then tests broke because it was throwing a `NullPointerException` in some cases

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CheckCollectionsForURI.java
ian-kent: I'm assuming it'll just do a string comparison against an empty string, which wouldn't match anything (since you (presumably) can't create content with an empty URI)

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
daiLlew: Probably because the test isn't configured to return anything. You can use Optional.empty() to create an empty instance of the optional.

PR #191, opened 2018-09-25
Improve slack notifications

ian-kent:
### What

Improves slack notifications - it was just a prototype, but it actually seems to work pretty well 👍 

* SlackNotification class has been cleaned up
* Messages use objects and JSON POST request instead of query parameters
* All notifications use slack attachments and fields to control layout
* Message presentation is more consistent
* Additional collection information is added to most alarms
* Publish log notifications are sent for all stages of the publish process
* Publish log notifications are updated in place with the latest data

### How to review

* Create a custom integration on slack to generate an API token
* Create a test channel
* Set `slack_default_channel`, `slack_alarm_channel`, `slack_publish_channel` and `slack_api_token` environment variables

#### Example tests

* Start zebedee, you should get the `Zebedee has just started` alarm
* Publish some manual and scheduled collections and check the publish log notifications
* Mark a collection as unencrypted then add content and publish it

### Who can review

Anyone (@daiLlew @mikeAdamss)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
```diff
@@ -164,8 +165,11 @@ public Boolean call() {
                 logError(e, "Exception saving collection after approval exception").collectionName(collection).log();
             }
 
-            SlackNotification.alarm(String.format("Exception approving collection %s : %s",
-                    collection.getDescription().getName(), e.getMessage()));
+            SlackNotification.collectionAlarm(collection,
+                    "Exception approving collection",
+                    new PostMessageField("Error", e.getMessage(), false)
+            );
```
daiLlew: Should this notification be inside the above catch block? Unless I am reading it incorrectly as it is it will always send this notification.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -27,88 +43,163 @@
  */
 public class SlackNotification {
 
+    public enum CollectionStage {
+        PrePublish,
```
daiLlew: Java naming convention for enum's is all caps.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -27,88 +43,163 @@
...
+    }
+
     private static final String slackToken = System.getenv("slack_api_token");
     private static final String slackDefaultChannel = System.getenv("slack_default_channel");
     private static final String slackAlarmChannel = System.getenv("slack_alarm_channel");
     private static final String slackPublishChannel = System.getenv("slack_publish_channel");
 
-    private static final String slackBaseUri = "https://slack.com/api/chat.postMessage";
-    private static final Host slackHost = new Host(slackBaseUri);
+    private static final String slackPostMessageURI = "https://slack.com/api/chat.postMessage";
```
daiLlew: I know it was already like this but if you are making changes to do it properly then we should probably make this config value.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -27,88 +43,163 @@
...
+    private static final String slackUpdateMessageURI = "https://slack.com/api/chat.update";
```
daiLlew: Same here

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -27,88 +43,163 @@
...
+
+    private static final String slackUsername = "Zebedee";
```
daiLlew: and here

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -27,88 +43,163 @@
...
-        alarm(msg.toString());
     }
 
-    public static void alarm(String message) {
-        String slackUsername = "Alarm";
-        String slackEmoji = ":heavy_exclamation_mark:";
-        send(message, slackToken, slackAlarmChannel, slackUsername, slackEmoji);
-    }
+    private static void addCollectionAttachment(PostMessageAttachment attch, CollectionBase collection) {
+        attch.fields.add(new PostMessageField("Collection", "<" + Configuration.getFlorenceUrl() + "/florence/collections/" + collection.getId() + "|" + collection.getName() + ">", true));
```
daiLlew: Appending these values inline feels bit nasty, IMO it would be better to have a format method and call that instead. it also makes reading this method easier as the input is shorter.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -27,88 +43,163 @@
...
-    public static void send(String message, String slackToken, String slackChannel, String slackUsername, String slackEmoji) {
+            if (publishedCollection.publishStartDate != null && publishedCollection.publishEndDate != null) {
+                String timeTaken = String.format("%.2f", (publishedCollection.publishEndDate.getTime()
+                        - publishedCollection.publishStartDate.getTime()) / 1000.0);
 
-        if (slackToken == null || slackChannel == null) {
-            return;
-        }
+                attch.fields.add(new PostMessageField("Duration", timeTaken + " seconds", true));
+                attch.fields.add(new PostMessageField("Publish start date", slackFieldFormatAccurate.format(publishedCollection.publishStartDate), true));
```
daiLlew: I would consider making "Publish start data" and the other Strings you are using a couple of times constants.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
```diff
@@ -117,52 +208,139 @@ public static void send(String message, String slackToken, String slackChannel,
...
-                    .filter(info -> info.uri.endsWith(DATA_JSON))
-                    .findFirst()
-                    .ifPresent(urlInfo -> {
-                        msg.append("Example Uri: http://www.ons.gov.uk")
-                                .append(urlInfo.uri.substring(0, urlInfo.uri.length() - (DATA_JSON).length()));
-                    });
-            slackMessage = msg.toString();
+        switch (stage) {
+            case PrePublish:
+                switch (status) {
```
daiLlew: [Personal opinion] I don't like the nested switch here I think it adds a lot of noise. I think it would be cleaner to have the top level switch here and then have separate methods for the stages. E.g.

```
switch (stage) {
            case PrePublish:
                prePublishNotification(...)
                break;
            case Publish:
                publishNotification(...)
                break;
....
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/PostMessage.java
```diff
@@ -0,0 +1,47 @@
+package com.github.onsdigital.zebedee.util.slack;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PostMessage {
+    public enum Emoji {
+        HeavyExclamationMark,
```
daiLlew: Naming convention for enum - upper case. Enums can have values e.g.
```
    public enum Emoji {
        HEAVY_EXCLAMATION_MARK(":heavy_exclamation_mark:"),

        CHART_WITH_UPWARDS_TREND(":chart_with_upwards_trend:");

        private final String icon;

        private Emoji(String icon) {
            this.icon = icon;
        }

        public String getIcon() {
            return icon;
        }
    }
```
then you can do `HeavyExclamationMark.getIcon()`

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/PostMessageAttachment.java
```diff
@@ -0,0 +1,42 @@
+package com.github.onsdigital.zebedee.util.slack;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PostMessageAttachment {
+    public enum Color {
+        Good,
```
daiLlew: Same here - upper case and consider adding the string value to the enum and using that in your constructor below.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/PostMessageAttachment.java
```diff
@@ -0,0 +1,42 @@
...
+        Warning,
+        Danger
+    }
+
+    public String pretext;
```
daiLlew: Unless there is a specific reason class vars should be private and you should use accessor methods to get/set them.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/PostMessage.java
```diff
@@ -0,0 +1,47 @@
...
+        ChartWithUpwardsTrend
+    }
+
+    public String channel;
```
daiLlew: should be private and have accessor methods

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/PostMessageField.java
```diff
@@ -0,0 +1,22 @@
+package com.github.onsdigital.zebedee.util.slack;
+
+import com.google.gson.annotations.SerializedName;
+
+public class PostMessageField {
+    public String title;
```
daiLlew: private and accessors...

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/PostMessageField.java
```diff
@@ -0,0 +1,22 @@
...
+    public String value;
+
+    @SerializedName("short")
+    public boolean isShort;
+
+    public PostMessageField(String title, String value) {
+        this.title = title;
+        this.value = value;
```
daiLlew: In this case `isShort`will have a default value of `false` - I assume that is OK? Also a common pattern for multiple constructors is to call the the "full" one passing in defaults for the missing values:
```
PostMessageField(String title, String value) {
    self(title, value, false)
}

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
ian-kent: It's inside the catch block, the inner catch block is handling cases where we set the approval status to error and fail to save - either way I'd assume we want to notify in slack?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
ian-kent: I'm not sure it's worth it for these, there's no alternative URLs we could ever use?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/SlackNotification.java
ian-kent: Same, the URL can never change

PR #220, opened 2018-12-11
Feature/approval error handling

daiLlew:
### What

- Refactor approval task to actually catch errors - `catch` now catches all exception instead of specific ones.
- Added new `ApprovalEventLog` to log the steps of approving to provide further logging details in error scenarios.
- Added new Event types `ApprovalSubmitted` and `ApprovalFailed` which is recorded in the collection json file.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
```diff
@@ -51,23 +56,133 @@
...
+            }
+            errorLog.log();
+            return false;
+        }
+    }
+
+    private boolean doApproval() throws Exception {
+        ApprovalEventLog eventLog = null;
+        try {
+            if (true) throw new RuntimeException("BAD STUFF HAPPENED");
```
daiLlew: FIX THIS

PR #243, opened 2019-02-04
Feature/upgrade logging collections history details

mikeAdamss:
### What

More api logging upgrades.

### How to review

Sanity check, try it.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
@@ -115,7 +115,7 @@ public class CollectionDetails {
         String collectionId = Collections.getCollectionId(request);
 
         if (datasetImportEnabled) {
-            info().data("collectionId", collectionId).data("user", session)
+            info().data("collectionId", collectionId).data("user", session.getEmail())
                 .log("CollectionDetails GET endpoint: datasetImportEnabled including dataset and dataset version " +
                         "details to response");
 
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
@@ -21,7 +21,7 @@ import javax.ws.rs.GET;
 import java.io.IOException;
 import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionDetails.java
```diff
@@ -112,9 +112,13 @@ public CollectionDetail get(HttpServletRequest request, HttpServletResponse resp
         result.teamsDetails = zebedeeCmsService.getZebedee().getTeamsService().resolveTeamDetails(teamIds);
         result.teamsDetails.forEach(team -> collection.getDescription().getTeams().add(team.getName()));
 
+        String collectionId = Collections.getCollectionId(request);
+
         if (datasetImportEnabled) {
-            logInfo("CollectionDetails GET endpoint: datasetImportEnabled including dataset and dataset version " +
-                    "details to response").collectionId(collection).user(session).log();
+            info().data("collectionId", collectionId).data("user", session)
```
daiLlew: `session.getEmail()`

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/CollectionHistory.java
```diff
@@ -60,9 +60,7 @@ private void checkPermission(Session session) throws ZebedeeException {
                 throw new UnauthorizedException("You are not authorised to create collections.");
             }
         } catch (IOException io) {
-            logError(io, "Unexpected error while trying to access permissions")
-                    .user(session.getEmail())
-                    .logAndThrow(UnexpectedErrorException.class);
+            error().data("user", session).logAndThrow (io, "Unexpected error while trying to access permissions");
```
daiLlew: `session.getEmail()`

daiLlew: I think this needs to have `throw` in front - the new log n throw method doesn't throw an exception is returns an exception so you have to declare a throw in order for it to happen... possibly a better name for this method is needed.

```
throw error().data("user", session).logAndThrow (io, "Unexpected error while trying to access permissions");
```

mikeAdamss: probably ok as you had it, this is more a bad assumption on my part that a code issue.

PR #251, opened 2019-02-11
initialise v2 DPlogger

mikeAdamss:
### What

Initialise the DPlogger.

### How to review

Sanity check.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
@@ -31,7 +31,7 @@ public class Init implements Startup {
         try {
             LogSerialiser serialiser = new JacksonLogSerialiser(true);
             LogStore logStore = new MDCLogStore(serialiser);
-            Logger logger = new LoggerImpl("com.test.app");
+            Logger logger = new LoggerImpl("com.zebedee.app");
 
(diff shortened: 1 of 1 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
```diff
@@ -19,7 +27,26 @@
 
     @Override
     public void init() {
-        DPLogger.init(new NopConfig());
+
+        try {
+            LogSerialiser serialiser = new JacksonLogSerialiser(true);
```
daiLlew: Remove the `true` from the constructor - for prod we don't want it to be formatted

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
```diff
@@ -19,7 +27,26 @@
...
+            LogStore logStore = new MDCLogStore(serialiser);
+            Logger logger = new LoggerImpl("com.zebedee.app");
```
daiLlew: This needs match the appender name in the logback.xml - `com.github.onsdigital.zebedee.cms`
https://github.com/ONSdigital/zebedee/blob/develop/zebedee-reader/src/main/resources/logback.xml.

PR #267, opened 2019-04-11
updated reader configuration to be somewhat sane and merged in featur…

daiLlew:
Intermitted test failures caused by system properties config values hanging around between tests. 
- Refactored reader configuration into a single place and hopefully made it sane.
- Deleted `ReaderFeatureFlags` and moved logic into `ReaderConfiguration` to eliminate circular dependency between them and to make it simpler and cleaner.
- Fixed any broken tests.


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/util/DatasetAPIClientSupplier.java
```diff
@@ -30,10 +27,9 @@ private static DatasetAPIClient getInstance() throws ZebedeeException {
                 if (INSTANCE == null) {
                     info().log("creating new instance of datasetAPIClient");
                     try {
-                        INSTANCE = new DatasetAPIClient(
-                                getDatasetAPIHost(),
-                                getDatasetAPIAuthToken(),
-                                getServiceAuthToken());
+                        ReaderConfiguration cgf = ReaderConfiguration.get();
```
ian-kent: typo in `cgf`?

PR #276, opened 2019-05-13
Feature/cmd dataset permissions service

daiLlew:
### What

Added new methods `getUserPermissions()`, `getServicePermissions` to the `AuthorisationService`. 

#### Admin/Editor users
If the user is an **admin/publisher** they will be granted the appropriate permissions if:
- Their session exists and is not expired
- The collection exists
- The collection contains the requested dataset id.
- They have`canEdit` CMS permissions.
##### Permissions granted
```
{ "permissions" : [ "CREATE", "READ", "UPDATE", "DELETE" ] }
```
#### Viewer users
If the user is a **viewer** they will be granted appropriate permissions if:
- Their session exists and is not expired
- The collection exists
- The collection contains the requested dataset id.
- **They are assigned to a team that is associated with the requested collection**
- They have`canView` CMS permissions.
##### Permissions granted
```
{ "permissions" : ["READ"] }
```
##### Default permissions granted
```
{ "permissions" : [ ] }
```

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/AuthorisationServiceImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/AuthorisationServiceImpl.java
@@ -83,25 +83,35 @@ public class AuthorisationServiceImpl implements AuthorisationService {
 
         validateCollectionContainsRequestedDataset(collection, datasetID);
 
-        DatasetPermissions permissions = new DatasetPermissions();
-
-        if (isAdminOrPublisher(session)) {
-            permissions.permit(CREATE, READ, UPDATE, DELETE);
-        } else if (isCollectionViewer(session, collection.getDescription())) {
-            permissions.permit(READ);
-        }
-
-        info().data("user", session.getEmail())
-                .data("collection_id", collectionID)
-                .data("dataset_id", datasetID)
-                .data("permissions", permissions.getPermissions())
-                .log("permitting dataset admit /editor permission");
-        return permissions;
+        return assignPermissions(session, collection.getDescription(), datasetID);
     }
 
     @Override
     public DatasetPermissions getServicePermissions(String serviceToken) throws DatasetPermissionsException {
         return null;
+    }
+
+    DatasetPermissions assignPermissions(Session session, CollectionDescription description, String datasetID)
+            throws DatasetPermissionsException {
+        DatasetPermissions permissions = new DatasetPermissions();
+
+        if (isAdminOrPublisher(session)) {
+            permissions.permit(CREATE, READ, UPDATE, DELETE);
+        } else if (isCollectionViewer(session, description)) {
+            permissions.permit(READ);
+        } else {
+            info().data("collection_id", description.getId())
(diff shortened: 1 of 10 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/AuthorisationServiceImpl.java
```diff
@@ -58,12 +71,131 @@ public UserIdentity identifyUser(String sessionID) throws UserIdentityException
...
     @Override
     public DatasetPermissions getServicePermissions(String serviceToken) throws DatasetPermissionsException {
         return null;
     }
+
+    DatasetPermissions assignPermissions(Session session, CollectionDescription description, String datasetID)
+            throws DatasetPermissionsException {
+        DatasetPermissions permissions = new DatasetPermissions();
+
+        if (isAdminOrPublisher(session)) {
```
mikeAdamss: will need to explain this one to me. Successful return is true? Error is false?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/AuthorisationServiceImpl.java
```diff
@@ -19,10 +29,13 @@
 
     private ServiceSupplier<SessionsService> sessionServiceSupplier = () -> Root.zebedee.getSessionsService();
     private ServiceSupplier<UsersService> userServiceSupplier = () -> Root.zebedee.getUsersService();
+    private ServiceSupplier<Collections> collectionsSupplier = () -> Root.zebedee.getCollections();
+    private ServiceSupplier<PermissionsService> permissionsServiceSupplier = () -> Root.zebedee.getPermissionsService();
 
     private static final String INTERNAL_ERROR = "internal server error";
     private static final String AUTHENTICATED_ERROR = "user not authenticated";
     private static final String USER_NOT_FOUND = "user does not exist";
+    private static final DatasetPermissions ADMIN_EDITOR_PERMISSIONS = new DatasetPermissions(CREATE, READ, UPDATE, DELETE);
```
mikeAdamss: cant see where this is used

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/authorisation/AuthorisationServiceImpl.java
daiLlew: If they are not an admin or publisher it's not a "failure" - they are still a valid user but they are a `viewer` and restricted in what they can see.

From the dataset permissions POV we treat admins and publishers the same - they are allowed to see all prerelease data so we grant them full `CRUD` permissions.

Viewer users have to be in a team which is assigned to a collection to be able to see it's content. So we have to do an addition check to determined if we should grant them any permissions at all. 
If they are in a team assigned to a collection we grant them `R` only, otherwise we grant nothing.

Both the `isAdminOrPublisher` and `isCollectionViewer` methods return `true` or `false` _but_ they can also throw an `IOException` - if this happens something has gone pretty wrong. In such a case we would bail out and return a `500`.

PR #293, opened 2019-08-15
Feature/update logger version

daiLlew:
### What

 - Update `dp-logger` version to latest release.
 - Updated `run.sh` to set `ENABLE_PERMISSIONS_AUTH=true` by default.
 - Removed last remnants of old logger impl and updated affected classes to use new world version.
 - Added `try-catch-log` in collection approve method to capture errors saving collection.
- Readability improvements to `Identity` api and better logging.
- Removed all service & session ids form logging.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/logging/ZebedeeLogBuilder.java
```diff
@@ -1,340 +0,0 @@
-package com.github.onsdigital.zebedee.logging;
```
daiLlew: Removed as it's legacy log stuff that is no longer required.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
```diff
@@ -99,7 +99,7 @@ public Boolean call() {
             return doApproval();
         } catch (Exception e) {
```
dandwal: We are not using the exception that is thrown but showing our own message. Would it be worth us logging the exact exception details thrown in addition to what we have? (Definitely not a blocker, as this PR doesn't really change the logic here). Same in a few other places

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
```diff
@@ -280,9 +280,16 @@ private void compressZipFiles(Collection collection, CollectionReader collection
...
     public void approveCollection() throws IOException {
         // set the approved state on the collection
-        collection.description.approvalStatus = ApprovalStatus.COMPLETE;
-        collection.description.addEvent(new Event(new Date(), EventType.APPROVED, session.getEmail()));
-        collection.save();
+        try {
+            collection.getDescription().setApprovalStatus(ApprovalStatus.COMPLETE);
+            collection.getDescription().addEvent(new Event(new Date(), EventType.APPROVED, session.getEmail()));
+            collection.save();
+        } catch (Exception ex) {
```
dandwal: Minor consistency comment most places we have been keeping exception name as `e` but here it is `ex`.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -98,26 +97,42 @@ private void findUser(HttpServletRequest request, HttpServletResponse response)
...
+        if (!authorizationHeader.startsWith("bearer ")) {
+            warn().log("authorization header does not begin with the required bearer prefix returning null");
+            return null;
+        }
+
+        String serviceToken = removeBearerPrefix(authorizationHeader);
+        ServiceAccount serviceAccount = null;
+
+        try {
+            serviceAccount = getServiceStore().get(serviceToken);
```
janderson2: On line 106 you are lowercasing the auth header therefore the service token has now been lowercased.  However from looking at the service store and service endpoint it does not look like we are enforcing that the token be lowercased there.  Since this is a direct lookup we probably want to be explicit about lowercasing everywhere.  Potentially it would be best to add this to the ServiceStoreImpl.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
daiLlew: By convention it should be `ex`.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
daiLlew: Existing logic. I haven't change it just rearranged it for readability.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
daiLlew: It's being added to the error log event on line 111 - its passed into `logException` - this adds the exception to the log event, sets the event message and logs it to std out.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
daiLlew: The lower casing is actually wrong and not needed. I can see why it was done originally - to make it easier to check for both `"Bearer"` and `"bearer"` however thats not correct. According to the specification the correct and expected value is `Bearer XXX`.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Identity.java
```diff
@@ -34,6 +33,8 @@
     private boolean datasetImportEnabled = false;
 
     static final String AUTHORIZATION_HEADER = "Authorization";
+    static final String BEARER_PREFIX_UC = "Bearer";
+    static final String BEARER_PREFIX_LC = "bearer";
```
janderson2: It doesn’t look like the lowercase version is being used anywhere (from what I can see on my phone anyway). Probably worth removing it for clarity. Could potentially remove the UC from the other contestant then.

PR #296, opened 2019-08-21
added regex validation for service token ids

daiLlew:
### Service token ID validation.

 - Remove enforced lower casing of token value when retrieving token from store. Fixes issue where token not found in case sensitive file systems. 
 - Improved general logging around service tokens to clearly communicate the reason when a token is not found.
- Added validation to service tokens
  - Tokens can only contain _alphanumeric_ characters.
  - Must have a length >= *16* characters.
 - Added new `ServiceTokenUtils` for service token validation and unit tests.
 - Updated `ServiceStoreImpl` to use `ServiceTokenUtils` for consistency and added unit tests as there were none.
- Updated `Identity` API to also use `ServiceTokenUtils` for validation of service token request header values.
- Added missing javadoc to various things because it was missing or incorrect. (Clean the campsite n all that)

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ServiceStoreImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ServiceStoreImpl.java
@@ -10,20 +10,33 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 
 import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
 
+/**
+ * File based service account store implementation.
+ */
 public class ServiceStoreImpl implements ServiceStore {
 
     private final Path serviceRoot;
 
     private final JSONSerialiser<ServiceAccount> jsonSerialise;
 
-    private final static String JSON_EXTENSION = ".json";
-
+    /**
+     * Construct a new ServiceStoreImpl instance.
+     *
+     * @param serviceRoot the file path of the service account root directory.
+     */
     public ServiceStoreImpl(Path serviceRoot) {
         this.serviceRoot = serviceRoot;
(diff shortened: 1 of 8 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ServiceStoreImpl.java
```diff
@@ -10,9 +10,14 @@
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.regex.Pattern;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
 
 public class ServiceStoreImpl implements ServiceStore {
 
+    private static final Pattern VALID_FILENAME_REGEX = Pattern.compile("^\\w+$");
```
gedge: Full disclosure - I suggested the `\w` used here, but that accepts underscores (as well as alphanumeric), which was not what Dave asked for - sorry.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ServiceStoreImpl.java
```diff
@@ -1,60 +1,79 @@
...
-        final Path path = getPath(id);
-        if (Files.exists(getPath(id))) {
-            try (InputStream input = Files.newInputStream(path)) {
-                service = jsonSerialise.deserialiseQuietly(input, path);
-            }
+    public ServiceAccount get(String token) throws IOException {
+        Path filePath = null;
+        ServiceAccount account = null;
+
+        if (ServiceTokenUtils.isValidServiceToken(token)) {
```
mikeAdamss: will we still have bearer prefix here?

daiLlew: No. The prefix should be removed by this point.

PR #297, opened 2019-09-02
Expose dataset information in publishedCollection data

dandwal:
### What

Added datasets and datasetVersions when being serialised

### How to review

Published a cmd dataset, and go to the publishedCollections or report screen, should see dataset information that was published e.g.: Screenshot 2019-09-02 at 16 47 53

### Who can review

Anyone except me


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/publishing/PublishedCollection.java
```diff
@@ -2,12 +2,19 @@
...
+import com.github.onsdigital.zebedee.json.CollectionDatasetVersion;
 
 import java.util.Date;
 import java.util.List;
+import java.util.Set;
+
 
 public class PublishedCollection extends CollectionBase {
 
+    public Set<CollectionDataset> datasets;
```
daiLlew: Java convention is that these should be private and there should be accessor methods. Other than than all good.

PR #309, opened 2019-11-28
code changes to clean up properly after deleting content

daiLlew:
#### The issue
When page is deleted from a collection the current implementation only deletes that page file. This is a problem if that page has created supplementary files - tables, charts, images etc. In this case only the page file (`data.json`) is deleted and the other files are left behind. 

These files are not visible or assessable to Florence users but prevent them from deleting files from the collection or the collection its self and/or from approving the collection.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
@@ -16,6 +16,7 @@ import com.github.onsdigital.zebedee.json.ApprovalStatus;
 import com.github.onsdigital.zebedee.json.Event;
 import com.github.onsdigital.zebedee.json.EventType;
 import com.github.onsdigital.zebedee.json.Keyring;
+import com.github.onsdigital.zebedee.logging.CMSLogEvent;
 import com.github.onsdigital.zebedee.model.approval.ApprovalQueue;
 import com.github.onsdigital.zebedee.model.approval.ApproveTask;
 import com.github.onsdigital.zebedee.model.publishing.PostPublisher;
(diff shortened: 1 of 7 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -734,7 +734,7 @@ public boolean deleteContent(
         } else if (Files.isDirectory(path)) {
             deleted = collection.deleteContentDirectory(session.getEmail(), uri);
         } else {
-            deleted = collection.deleteFile(uri);
+            deleted = deleteFile(collection, uri);
```
jessjenkins: Method name is slightly misleading, it's not deleting just a file, it's deleting more than that. Not sure exactly what's best but something (?) along the lines of deleteFileAndRelated()

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Content.java
```diff
@@ -375,6 +379,50 @@ public boolean delete(String uri) throws IOException {
...
+        }
+
+        return deleteSuccessful;
+    }
+
+    /**
+     * Delete files from the specified collection directory. directories and cy_data.json are exluded all other files
+     * will be deleted.
+     */
+    boolean deleteDataJsonFile(Path path) throws IOException {
```
jessjenkins: Method name again, this method is deleting a directory containing the data json, not just a file.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -717,34 +718,45 @@ public boolean deleteContent(
...
+    boolean deletcContentFromCollection(Collection collection, Session session, Path contentTargetPath, String uri) throws IOException {
         boolean deleted;
-        CollectionEventType eventType;
+        Path uriPath = Paths.get(uri);
+
+        CMSLogEvent event = info().collectionID(collection).uri(contentTargetPath);
 
-        if (isDataVisualisationFile(path)) {
+        if (isDataVisualisationFile(contentTargetPath)) {
+            event.log("deleting data viz zip form collection content");
```
eldeal:
```suggestion
            event.log("deleting data viz zip from collection content");
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -717,34 +718,45 @@ public boolean deleteContent(
...
             deleted = collection.deleteDataVisContent(session, Paths.get(uri));
-        } else if (Files.isDirectory(path)) {
+        } else if (Files.isDirectory(contentTargetPath)) {
+            event.log("deleting directory form collection content");
```
eldeal:
```suggestion
            event.log("deleting directory from collection content");
```

PR #316, opened 2020-01-15
made collection content fields private and added accessors to follow …

daiLlew:
Changes collection content fields to private and added accessor methods to follow good java developing standards and updated all uses to use appropriate method.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1481,5 +1476,17 @@ private static String collectionContentPath(String collectioName, String uri) {
...
+    public Content getReviewed() {
+        return this.reviewed;
+    }
+
+    public Content getComplete() {
+        return this.complete;
+    }
+
+    public Content getInProgress() {
+        return this.complete;
```
rav-pradhan: Should this return `inProgress`?

PR #320, opened 2020-01-30
Added /health endpoint (always returns OK), and whitelisted it for authentication

DavidSubiros:
### What

- Added Health API class, which supports /GET operations and returns OK
- Added Health endpoint to NO_AUTH_REQUIRED in AuthenticationFilter

### How to review

- Make sure the changes make sense
- You can try to run Zebedee, and perform a GET against the new endpoint. It should return 200 OK with an empty body.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Health.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Health.java
@@ -1,0 +1,20 @@
+package com.github.onsdigital.zebedee.api;
+
+import com.github.davidcarboni.restolino.framework.Api;
+import org.apache.http.HttpStatus;
+
+import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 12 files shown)


DavidSubiros:
@eldeal  I didn't find any health-check endpoint previously, that's why we decided to just create a new one and put it to the NO_AUTH_REQUIRED list. 
The behaviour we saw originally (401 response) happens for any path that does not exist.

DavidSubiros:
rebased

nshumoogum:
> Do we need to add the previous healthcheck endpoint to the NOAUTHREQUIRED list as well?

No because the `/healthcheck` endpoint did not previously exist and therefore if any application needed this endpoint then it would be currently broken; this is clearly not the case otherwise it would be a priority to fix it.

When implementing a checker for any service that relies on zebedee, it will use the `/health` endpoint that David has implemented here. I hope that clears it up

PR #322, opened 2020-02-04
set 3rd party log events to level WARN to reduce unncessary spam

daiLlew:
Changed 3rd party log event level to `WARN` to reduce unnecessary logging output.

### DIFF
--- a/zebedee-reader/src/main/resources/logback.xml
+++ b/zebedee-reader/src/main/resources/logback.xml
@@ -45,7 +45,7 @@
     </logger>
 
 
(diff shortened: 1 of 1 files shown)


zebedee-reader/src/main/resources/logback.xml
```diff
@@ -45,7 +45,7 @@
     </logger>
 
 
-    <root level="info">
+    <root level="WARN">
```
jessjenkins: Assume the capitalisation is OK.

PR #326, opened 2020-02-10
removed redundant logging calls and moved health check from CMS to re…

daiLlew:
#### Fix healthcheck
- New Health check api was added to Zebedee CMS so was only ever available when running in CMS mode so moved into Zebedee reader - it will be available in both reader and CMS.

### Logging
- Removed redundant logging calls. Logs for service accounts and Identity are very noisy. Removed log events adding little/no value.


DavidSubiros:
thanks for this change, it looks good to me too.

PR #337, opened 2020-03-20
fixed missing version event

daiLlew:
### What
Fixed defect where add/delete content version events were not being persisted to the collection json.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Version.java
```diff
@@ -77,7 +88,12 @@ public boolean delete(HttpServletRequest request, HttpServletResponse response)
         Collection collection = Collections.getCollection(request);
         String uri = request.getParameter("uri");
 
-        collection.deleteVersion(uri);
+        try {
+            info().uri(uri).collectionID(collection).user(session).log("deleting colleciton version content");
```
CarlHembrough: typo in log message

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1228,7 +1229,7 @@ public ContentItemVersion version(String email, String uri, CollectionWriter col
      * @throws NotFoundException   - if the given URI was not found in the collection.
      * @throws BadRequestException - if the given URI is not a valid version URI.
      */
-    public void deleteVersion(String uri) throws NotFoundException, BadRequestException, IOException {
+    public void deleteVersion(String email, String uri) throws NotFoundException, BadRequestException, IOException {
```
CarlHembrough: Consider adding the new parameter to the Java doc above the method

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/content/item/VersionedContentItem.java
```diff
@@ -29,6 +33,7 @@
 
     static final String VERSION_DIRECTORY = "previous";
     static final String VERSION_PREFIX = "v";
+    static final Pattern VERSION_DIR_PATTERN = Pattern.compile("v\\d+");
```
CarlHembrough: Consider adding the 'previous' directory into this regular expression, to safeguard against any page titles that have `v..` in them.

e.g. this test would fail
```
@Test
    public void getVersionFromURI_uriForVersionFile_shouldReturnVersion() {
        Optional<String> result = VersionedContentItem.getVersionFromURI("/a/bv2/veryimportantpage/previous/v1/data.json");
        assertTrue(result.isPresent());
        assertThat(result.get(), equalTo("v1"));
    }
```

The matcher could be updated to something like: `VERSION_DIR_PATTERN = Pattern.compile("previous/(v\\d+)");`
(also would require the matcher group to be updated to 1 `result = Optional.of(matcher.group(1));`)

PR #338, opened 2020-03-24
Feature/review dataset warning

daiLlew:
### Defect #4687

Still looking for a cause for this defect. Added check to prevent a collection being approved if the expected dataset versions are missing.

Added override mechanism for desperate situations.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -75,9 +75,10 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.error;
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.info;
 import static com.github.onsdigital.zebedee.configuration.CMSFeatureFlags.cmsFeatureFlags;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
 import static com.github.onsdigital.zebedee.model.content.item.VersionedContentItem.getVersionFromURI;
 import static com.github.onsdigital.zebedee.model.content.item.VersionedContentItem.isVersionedUri;
 import static com.github.onsdigital.zebedee.persistence.CollectionEventType.COLLECTION_CONTENT_REVIEWED;
@@ -847,12 +848,9 @@ public class Collection {
             Path destination = reviewed.toPath(uri);
 
             if (recursive) {
-                FileUtils.deleteDirectory(destination.getParent().toFile());
-                FileUtils.moveDirectory(source.getParent().toFile(), destination.getParent().toFile());
-                zebedee.getCollections().removeEmptyCollectionDirectories(source.getParent());
+                reviewRecursive(source, destination, session);
             } else {
-                PathUtils.moveFilesInDirectory(source, destination);
-                zebedee.getCollections().removeEmptyCollectionDirectories(source);
+                reviewSingleFile(source, destination);
             }
 
             addEvent(uri, new Event(new Date(), EventType.REVIEWED, session.getEmail()));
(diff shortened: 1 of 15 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -865,6 +867,27 @@ public boolean review(Session session, String uri, boolean recursive) throws IOE
         return result;
     }
 
+    private void reviewRecursive(Path src, Path dest, Session session) throws IOException {
+        if (src.toString().contains("/dataset/")) {
```
CarlHembrough: It looks like `/dataset/` should be plural `/datasets/` ?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -964,4 +975,20 @@ public boolean hasCollection(String name) {
...
     }
+
+    public void verifyDatasetVersions(Collection collection, CollectionReader collectionReader, Session session) throws ZebedeeException, IOException {
+        if (CMSFeatureFlags.cmsFeatureFlags().isDatasetVersionVerificationEnabled()) {
+            info().collectionID(collection).log("dataset version verification feature enabled verifying collection");
+
+            try {
+                versionsService.verifyCollectionDatasets(zebedeeCmsService.getZebedeeReader(), collection, collectionReader, session);
+            } catch (VersionNotFoundException ex) {
+                info().collectionID(collection)
```
CarlHembrough: should this be `error()` instead of `info()`?

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/statistics/dataset/Version.java
```diff
@@ -1,6 +1,7 @@
 package com.github.onsdigital.zebedee.content.page.statistics.dataset;
 
 import java.net.URI;
+import java.util.Comparator;
```
CarlHembrough: Remove if not needed

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/versioning/VersionsServiceImpl.java
```diff
@@ -0,0 +1,177 @@
...
+                    .log("collection legacy datasets verified successfully");
+        }
+    }
+
+    List<Dataset> getDatasetsInCollection(CollectionReader reader) throws ZebedeeException, IOException {
+        List<Dataset> datasets = new ArrayList<>();
+
+        for (String uri : reader.getReviewed().listUris()) {
+
+            if (isDataJson(uri) && !isVersionedURI(uri)) {
```
CarlHembrough: consider filtering by pages with `/datasets/` in the URI. That way it will not load the content for all data.json files in the collection. This could be quite expensive for large collections

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/versioning/VersionsServiceImpl.java
```diff
@@ -0,0 +1,177 @@
...
+        for (Dataset ds : datasets) {
+            info().collectionID(collection)
+                    .data("dataset", ds.getDescription().getTitle())
+                    .log("verifying collection dataset versions");
+
+            for (Version version : ds.getVersions()) {
+                String uri = version.getUri().toString();
+
+                try {
+                    cmsReader.getCollectionContent(collection.getId(), session.getId(), uri);
```
CarlHembrough: Consider checking also if the excel file exists for the dataset if its found in the collection

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/versioning/VersionsServiceImpl.java
```diff
@@ -0,0 +1,177 @@
...
+
+    void verifyDatasetVersions(ZebedeeReader cmsReader, Collection collection, Session session,
+                               List<Dataset> datasets) throws ZebedeeException, IOException,
+            VersionNotFoundException {
+        for (Dataset ds : datasets) {
+            info().collectionID(collection)
+                    .data("dataset", ds.getDescription().getTitle())
+                    .log("verifying collection dataset versions");
+
+            for (Version version : ds.getVersions()) {
```
CarlHembrough: Is it always the case that the version is added to the data.json? If not, consider checking for the existence of any version of a dataset when a dataset is in a collection

eldeal: Would every version exist in the collection? or just the latest version? Not sure if this should be a loop or just like a 'find the latest' and then check the collection content 🤔

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/util/versioning/VersionsServiceImplTest.java
```diff
@@ -0,0 +1,281 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.when;
+
+public class VersionsServiceImplTest {
+
+    static final String CONTENT_PATH = "/a/b/c/data.json";
+    static final String CONTENT_URI = "/a/b/c";
+    static final String VERSOIN_URI = CONTENT_URI + "/previous/v1";
```
eldeal: We got a little typo here on the VERSION_URI var name (and throughout this file)

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/versioning/VersionsServiceImpl.java
CarlHembrough: Only the latest version should be in the collection. I was just wondering if the data.json was not updated with the new version, how would we verify that.

PR #342, opened 2020-03-31
Add dependency audit and speed up reader builds

janderson2:
### What

Add dependency vulnerability audit to the build and speed up the reader
CI jobs by excluding zebedee-cms.

### Who can review

Anyone but me.

### DIFF
--- a/ci/scripts/build.sh
+++ b/ci/scripts/build.sh
@@ -1,11 +1,11 @@
 #!/bin/bash -eux
 
 pushd zebedee
-    mvn -DskipTests=true clean package dependency:copy-dependencies
-
     if [[ "$APPLICATION" == "zebedee" ]]; then
+        make build-cms
         cp -r Dockerfile.concourse zebedee-cms/target/* ../build
     elif [[ "$APPLICATION" == "zebedee-reader" ]]; then
+        make build-reader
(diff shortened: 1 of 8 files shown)


ci/scripts/build.sh
```diff
@@ -1,11 +1,11 @@
 #!/bin/bash -eux
 
 pushd zebedee
-    mvn -DskipTests=true clean package dependency:copy-dependencies
-
     if [[ "$APPLICATION" == "zebedee" ]]; then
+        mvn -D skipTests=true -D ossindex.skip=true clean package dependency:copy-dependencies
```
daiLlew, line 5: I think the space between `-D` and the property is not needed i.e it should be
```
-Dossindex.skip=true
```

ci/scripts/build.sh
```diff
@@ -1,11 +1,11 @@
...
         cp -r Dockerfile.concourse zebedee-cms/target/* ../build
     elif [[ "$APPLICATION" == "zebedee-reader" ]]; then
+        mvn -pl zebedee-reader -D skipTests=true -D ossindex.skip=true clean package dependency:copy-dependencies
```
daiLlew, line 8: Ditto

PR #343, opened 2020-04-02
refactor error message and added more unit tests

daiLlew:
### What

Added some more unit tests for dataset version approval override.

Update readme.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Approve.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Approve.java
@@ -6,6 +6,7 @@ import com.github.onsdigital.zebedee.exceptions.NotFoundException;
 import com.github.onsdigital.zebedee.exceptions.UnauthorizedException;
 import com.github.onsdigital.zebedee.exceptions.ZebedeeException;
 import com.github.onsdigital.zebedee.session.model.Session;
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jetty.http.HttpStatus;
 
 import javax.servlet.http.HttpServletRequest;
@@ -22,6 +23,8 @@ import static com.github.onsdigital.logging.v2.event.SimpleEvent.warn;
  */
 @Api
 public class Approve {
+
+    static final String OVERRIDE_KEY_PARAM = "overrideKey";
 
     /**
      * Approves the content of a collection at the endpoint /Approve/[CollectionName].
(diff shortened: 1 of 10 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Approve.java
```diff
@@ -76,13 +79,24 @@ public boolean approveCollection(HttpServletRequest request, HttpServletResponse
...
+        if (StringUtils.isEmpty(overrideKey)) {
+            return null;
+        }
+
+        Long key = null;
         try {
             key = Long.valueOf(overrideKey);
         } catch (NumberFormatException ex) {
-            // use default
+            info().exception(ex).log("failed to convert override parameter to long");
```
CarlHembrough, line 99: should this be error() ?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/versioning/VersionNotFoundException.java
```diff
@@ -24,16 +22,15 @@ public VersionNotFoundException(final Throwable cause) {
...
 
-    public static VersionNotFoundException versionsNotFoundException(Collection collection,
-                                                                     List<MissingVersion> missingVersions) {
+    public static VersionNotFoundException versionsNotFoundException(List<MissingVersion> missingVersions) {
         String details = missingVersions
                 .stream()
-                .map(i -> i.toString())
+                .map(i -> i.getCurrentURI())
+                .collect(Collectors.toSet())
+                .stream()
```
CarlHembrough, line 30: Why is this being collected as a set and then streamed again? I could only think it was to de-dupe the collection. If so, consider using `.distinct()` to save collecting and streaming twice

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Approve.java
daiLlew: No. In this case it would continue with null. I added the log just for visibility

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/versioning/VersionNotFoundException.java
daiLlew: 100%. I was looking for something to de-dupe but couldn't think what it was called. I will make the change now.

PR #344, opened 2020-04-02
Update cryptolite dependency to use ONS fork

jessjenkins:
### What

Updated the cryptolite dependency to use the ONS hosted fork.

### How to review

Ensure zebedee builds and runs without changes. 

### Who can review

Anyone but me but preferably dave-Llew and janderson2

### DIFF
--- a/zebedee-cms/pom.xml
+++ b/zebedee-cms/pom.xml
@@ -41,8 +41,8 @@
             <artifactId>restolino</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.github.davidcarboni</groupId>
-            <artifactId>cryptolite</artifactId>
+            <groupId>com.github.onsdigital</groupId>
+            <artifactId>dp-cryptolite-java</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


zebedee-cms/pom.xml
```diff
@@ -67,6 +67,14 @@
             <groupId>com.github.davidcarboni</groupId>
             <artifactId>encrypted-file-upload</artifactId>
             <version>1.0.0</version>
+            <exclusions>
```
daiLlew, line 70: Seems like a sensible approach. Only question I have is are sure that this still works correctly using the newer version - just checking? Also are we considering forking this as well so we can update the dependency or make it `provided`?

jessjenkins: I have tested extensively locally and this does work. I also intend to fully test in the develop environment before release. There was only a small change between v1.3.0 and v1.3.2 and it was already impacted by this as the existing pom ignored 1.3.0 anyway due to the conflict.
On the second point, yes we really should fork `encrypted-file-upload` but I think this is best done as a separate 20% task.

daiLlew: Good stuff. I assumed you had tested it but it wouldn't be a worthwhile review if I didn't ask 😄

PR #346, opened 2020-04-08
fixed incorrect logic in remove previous versions code.

daiLlew:
Fix for defect where all dataset previous versions were being deleted from the collection if a dataset page or dataset landing page were deleted.

Filter logic was using incorrect variable was evaluating to true for all versioned URIs instead of only returning versioned URIs that are children of the target URI.

🤦‍♂

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -1056,7 +1056,7 @@ public class Collection {
      */
     private boolean deletePreviousVersionFromReviewed(String targetURI) throws IOException {
         boolean deleteSuccessful = true;
-        List<String> versionedFiles = getPreviousVersionContentForURI(reviewedUris(), targetURI);
+        List<String> versionedFiles = versionsService.getPreviousVersionsOf(targetURI, reviewed);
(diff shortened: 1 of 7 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1051,20 +1050,19 @@ public boolean deleteFileAndRelated(String uri) throws IOException {
...
      * @return
      * @throws IOException
      */
-    private boolean deleteRelatedVersionedContent(String uri) throws IOException {
+    private boolean deletePreviousVersionFromReviewed(String targetURI) throws IOException {
         boolean deleteSuccessful = true;
-
-        List<String> versionedFiles = reviewedUris()
-                .stream()
-                .filter(contentURI -> contentURI.startsWith(contentURI) && isVersionedUri(contentURI))
```
daiLlew, line 1063: This is the origin of the defect.

It should be iterating through the collection reviewed content and finding any files that start with the URI of the file being deleted  `AND` are a versioned URI.

However this mistake is comparing the uri of each reviewed filed to its self. This will always evaluate `true` the result is it matches all previous version files in the reviewed instead of just the versions of the file being deleted.

PR #353, opened 2020-05-05
Upgrade spring dependency

jessjenkins:
### What

Upgraded spring dependency to 5.2.5 (latest)
This is mainly used for testing, although there was one main usage (TeamsReport) where it was being used for a couple of constants. I switched those constants to ones from other libraries to be consistent with other classes in the app. This allowed me to scope spring to just the test lifecycle. 
 
### How to review

Ensure zebedee builds, tests and runs successfully.
Ensure spring dependencies are of version 5.2.5 (5.1.1 or later required for security auditing resolution)

### Who can review

Anyone but me

### DIFF
--- a/zebedee-cms/pom.xml
+++ b/zebedee-cms/pom.xml
@@ -30,7 +30,7 @@
     </repositories>
 
     <properties>
-        <spring.version>4.2.6.RELEASE</spring.version>
+        <spring.version>5.2.5.RELEASE</spring.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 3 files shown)


zebedee-cms/pom.xml
```diff
@@ -173,7 +174,8 @@
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-web</artifactId>
-            <version>4.3.8.RELEASE</version>
+            <version>${spring.version}</version>
+            <scope>test</scope>
```
daiLlew, line 178: I think this dependency can actually be removed now. I had a look and it appears this was the only place using it.

jessjenkins: Thanks for fixing the test. This is now merged into this branch. I approve your changes ✅.

PR #358, opened 2020-05-18
Add featured content to homepage model

jondewijones:
### What

- Add feature content property to homepage model
- Add reusable image model
- Add home content item type for use as featured content on homepage

### How to review

Sanity check

### Who can review

Anyone

### DIFF
--- a/zebedee-cms/pom.xml
+++ b/zebedee-cms/pom.xml
@@ -30,7 +30,7 @@
     </repositories>
 
     <properties>
-        <spring.version>4.2.6.RELEASE</spring.version>
+        <spring.version>5.2.5.RELEASE</spring.version>
     </properties>
 
     <dependencies>
@@ -144,6 +144,7 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-test</artifactId>
             <version>${spring.version}</version>
+            <scope>test</scope>
         </dependency>
 
         <dependency>
@@ -173,7 +174,8 @@
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-web</artifactId>
-            <version>4.3.8.RELEASE</version>
+            <version>${spring.version}</version>
+            <scope>test</scope>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/home/HomePage.java
```diff
@@ -40,6 +41,14 @@ public void setSections(List<HomeSection> sections) {
         this.sections = sections;
     }
 
+    public List<HomeContentItem> getFeaturedContent() {
```
daiLlew, line 44: Its probably advisable to initialise these fields in the constructor so they are not null.
```
private List<HomeSection> sections;
private List<HomeContentItem> featuredContent;
```
Something like
```
this.sections = new ArrayList<>();
this.featuredContent = new ArrayList():
```

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/home/HomeContentItem.java
```diff
@@ -0,0 +1,38 @@
...
+import com.github.onsdigital.zebedee.content.partial.Image;
+
+import java.net.URI;
+
+public class HomeContentItem {
+
+    private String title;
+    private URI uri;
+    private String description;
+    private Image image;
```
daiLlew, line 12: Add 2 constructors 1 which is empty, the 2nd taking these fields and setting them on the object. Makes it nicer for someone else to use.

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/partial/Image.java
```diff
@@ -0,0 +1,17 @@
+package com.github.onsdigital.zebedee.content.partial;
+
+import java.net.URI;
+
+public class Image {
+
+    private URI uri;
```
daiLlew, line 7: Add a constructor accepting the URI as a param.

PR #368, opened 2020-12-02
fixed content type header and added unit tests

daiLlew:
### What

Fix for defect in detecting file `MIME` types.

Existing logic to determined the `MIME` type for requested content was always returning null and defaulting to `application/octet-stream`. This was breaking data visualisations as the css files were not being returned with the correct content type header to be recognised.

The `Files.probeContentType()` method in the Java standard lib is apparent buggy and OS dependant 
https://stackoverflow.com/questions/12407479/why-does-files-probecontenttype-return-null

I've updated it to use `apache Tika` and added unit tests to prove it is resolving the types corrently.

### How to review

PR

### Who can review

Anyone.

### DIFF
--- a/zebedee-reader/pom.xml
+++ b/zebedee-reader/pom.xml
@@ -113,6 +113,12 @@
             <version>${fasterxml.jackson.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.tika</groupId>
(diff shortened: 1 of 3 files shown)


dandwal (dismissed):
I have tested a vis on IE, Chrome and Firefox all seem happy with this.

dandwal (approved):

PR #370, opened 2020-12-16
Add PublishingQueueCollectionDetails endpoint

CarlHembrough:
### What
A clone of the existing collection details endpoint, but without populating any lists of files that are in the collection. This is due to the issue of having too many files listed in the publish queue screen, which prevents it from loading.

Note: this PR removes *all* listing of files, including datasets. I thought it best to be consistent and either list all files or not.

### How to review
Sanity check

Test that the publishing queue loads correctly if you have the publishing system running locally.

### Who can review
Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/PublishingQueueCollectionDetails.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/PublishingQueueCollectionDetails.java
@@ -1,0 +1,60 @@
+package com.github.onsdigital.zebedee.api;
+
+import com.github.davidcarboni.restolino.framework.Api;
+import com.github.onsdigital.zebedee.exceptions.ZebedeeException;
+import com.github.onsdigital.zebedee.json.CollectionDetail;
+import com.github.onsdigital.zebedee.session.model.Session;
+import com.github.onsdigital.zebedee.util.ZebedeeCmsService;
+import org.eclipse.jetty.http.HttpStatus;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/PublishingQueueCollectionDetails.java
```diff
@@ -0,0 +1,60 @@
...
+     *                 <li>If no collection exists:  {@link HttpStatus#NOT_FOUND_404}</li>
+     *                 </ul>
+     * @return the CollectionDetail.
+     * @throws IOException
+     */
+    @GET
+    public CollectionDetail get(HttpServletRequest request, HttpServletResponse response)
+            throws IOException, ZebedeeException {
+
+        com.github.onsdigital.zebedee.model.Collection collection = Collections
```
rav-pradhan, line 34: Is there a reason we're not importing this like the statements at the top of the file?

rav-pradhan (commented):
I was able to test locally and could see the endpoint returning the relevant data in Florence. Left a minor comment but that wouldn't block approval.

Not sure if the policy is different for legacy systems but should we still add a test suite to cover the use case?
Do we want to add a test suite for this?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/PublishingQueueCollectionDetails.java
CarlHembrough: I think it has been done elsewhere due to name clashes, where the same class name is used in two different namespaces. I don't think it's required here though, so I will do the import 👍

rav-pradhan (approved):

PR #371, opened 2021-01-07
Start of integration of sessions api

morses-code:
- Added java client to the POM
- Started implementing the service


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -128,7 +133,9 @@ public ZebedeeConfiguration(Path rootPath, boolean enableVerificationAgent) thro
         this.dataIndex = new DataIndex(new FileSystemContentReader(publishedContentPath));
         this.publishedCollections = new PublishedCollections(publishedCollectionsPath);
         this.applicationKeys = new ApplicationKeys(applicationKeysPath);
-        this.sessions = new SessionsServiceImpl(sessionsPath);
+        this.sessionClient = new SessionClientImpl(Configuration.getSessionsApiUrl(), Configuration.getServiceAuthToken(), new Http());
+        //this.sessions = new SessionsServiceImpl(sessionsPath);
```
DavidSubiros, line 137: I think this commented line could be removed

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Check if the provided {@link Session} is expired.
+     *
+     * @param session the {@link Session} to check.
+     * @return true if expired, false otherwise.
+     */
+    @Override
+    public boolean expired(Session session) {
+        return false;
```
DavidSubiros, line 105: I guess the expiry will be implemented in the future, is it correct?

DavidSubiros (commented):
I failed to run the tests locally, but it is probably something wrong with my environment, as they run correctly in the CI.

Just added a minor comment and a question, the code looks good.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
morses-code: Expiry is all handled in the sessions api, I'm pretty sure I can remove this.

DavidSubiros (approved):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Check if an active {@link Session} exists with the provided ID.
+     *
+     * @param id the session ID to find.
+     * @return true if an active session exists with this ID, false otherwise.
+     * @throws IOException problem checking the session exists.
+     */
+    @Override
+    public boolean exists(String id) throws IOException {
+        return false;
```
daiLlew, line 94: Is this meant to return a hardcoded `false`? If I remember correctly this functionality is legacy and not actually needed. If that is the case I would personally update this to always throw a `UnsupportedOperationException`. This will make it very obvious this method is being used when it should be, returning false just masks that its being used.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
daiLlew: Am I right in saying this is a legacy method that is no longer needed? If so update it to always throw an `UnsupportedOperationException` just makes it clear and obvious its being called when it shouldn't be. Once we've migrated completely to the new sessions service we can consider deleting the method from the interface, technical that is naughty but its only used within Zebedee so it's unlikely to affect anyone but us.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Get the expiry date of the provided {@link Session}
+     *
+     * @param session the  {@link Session} to use.
+     * @return the sessions expiration date time as a {@link Date} instance.
+     */
+    @Override
+    public Date getExpiryDate(Session session) {
+        return null;
```
daiLlew, line 116: Same here - UnsupportedOperationException

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private Session createZebedeeSession(com.session.service.Session clientSession) {
+        Session session = new Session();
+        session.setId(clientSession.getId());
+        session.setEmail(clientSession.getEmail());
```
daiLlew, line 122: Is there any checking that the session from client is valid? I think we should be checking the ID and email are not null - if there are we should throw an ex.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/ZebedeeTestBaseFixture.java
```diff
@@ -78,10 +84,20 @@ public void init() throws Exception {
...
 
+        Session session = new Session();
+        session.setEmail(builder.publisher1.getEmail());
+        session.setId("1234");
+        session.setLastAccess(new Date());
+        session.setStart(new Date());
+
+        when(newSessionsService.create(any(User.class))).thenReturn(session);
+        when(newSessionsService.get(anyString())).thenReturn(session);
+        when(newSessionsService.find(anyString())).thenReturn(session);
```
daiLlew, line 95: Nit picky but this should probably specify the string value to return a session for:

````
when(newSessionsService.find(eq("1234")).thenReturn(session);
````
It ensures the test is providing the expected session to this mock call.

File: zebedee-cms/src/test/java/com/github/onsdigital/zebedee/ZebedeeTestBaseFixture.java [hunk 7]
```diff
@@ -78,10 +84,20 @@ public void init() throws Exception {
...
                 .thenReturn(usersList);
 
+        Session session = new Session();
+        session.setEmail(builder.publisher1.getEmail());
+        session.setId("1234");
+        session.setLastAccess(new Date());
+        session.setStart(new Date());
+
+        when(newSessionsService.create(any(User.class))).thenReturn(session);
+        when(newSessionsService.get(anyString())).thenReturn(session);
```
daiLlew, line 94: Same here.

daiLlew (changes requested):

zebedee-cms/pom.xml
```diff
@@ -191,6 +191,12 @@
             <version>${fasterxml.jackson.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.github.onsdigital</groupId>
+            <artifactId>dp-session-service-client-java</artifactId>
+            <version>0.1.1</version>
```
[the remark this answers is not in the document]
morses-code: Having a look at this, looks like the `dp-sessions-serivce-java-client` has `1.0-SNAPSHOT` version. So have changed it to that for now

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
morses-code: Makes sense, have updated it

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
morses-code: Added some checking on the id and email values that the client returns.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/ZebedeeTestBaseFixture.java
morses-code: There is one test that breaks when I do this and I can't figure out how to fix it. Not sure if you want to take a look?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/ZebedeeTestBaseFixture.java
morses-code: Same as mentioned above

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
morses-code: Done

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
morses-code: Done

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    public Session create(User user) throws IOException {
+        SessionCreated sessionCreated = client.createNewSession(user.getEmail());
+        com.session.service.Session clientSession = client.getSessionByID(sessionCreated.getId());
+
+        if (clientSession == null) {
+            throw new IOException();
+        }
+
+        Session session = createZebedeeSession(clientSession);
```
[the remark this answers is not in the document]
morses-code: I think initially I had more code here! Thanks

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+     * @return a {@link Session} instance for the user.
+     * @throws IOException problem creating a session.
+     */
+    @Override
+    public Session create(User user) throws IOException {
+        SessionCreated sessionCreated = client.createNewSession(user.getEmail());
+        com.session.service.Session clientSession = client.getSessionByID(sessionCreated.getId());
+
+        if (clientSession == null) {
+            throw new IOException();
```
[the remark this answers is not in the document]
morses-code: Added infomation for the error

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+import com.session.service.entities.SessionCreated;
+import com.session.service.error.SessionClientException;
+import org.apache.commons.lang.StringUtils;
+
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.util.Date;
+
+public class NewSessionsServiceImpl implements Sessions {
```
daiLlew, line 17: As this is a new class you should write the appropriate Javadoc on the class and the public methods. It doesn't have to be an essay just a succinct description of what this class does + any useful information anyone using it will need to know.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private SessionClient client;
+
+    public NewSessionsServiceImpl(SessionClient sessionClient) {
```
daiLlew, line 21: Document the constructor.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Get a {@link Session} from the provided {@link HttpServletRequest}.
+     *
+     * @param request the {@link HttpServletRequest} to get the session for.
+     * @return a {@link Session} instance if once exists returns null otherwise.
+     * @throws IOException for any problem getting a sesison from the request.
+     */
+    @Override
+    public Session get(HttpServletRequest request) throws IOException {
+        String token = RequestUtils.getSessionId(request);
```
daiLlew, line 54: Minor comment should rename this to `ID` or `sessionID` to be consistent with the rest of this class.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Check if an active {@link Session} exists with the provided ID.
+     *
+     * @param id the session ID to find.
+     * @return true if an active session exists with this ID, false otherwise.
+     * @throws IOException problem checking the session exists.
+     */
+    @Override
+    public boolean exists(String id) throws IOException {
+        throw new UnsupportedOperationException();
```
daiLlew, line 96: Can you update the javadoc for this method to say that its a deprecated method and not supported by this implementation add a message to this exception to give the error come context.

````
throw new UnsupportedOperationException("exists is a deprecated method and not supported by this sessions implementation.");
````
Also for extra brownie points can you add the `@Deprecated` annotation to this method declaration in the `Sessions` interface? This will indicate to any future users that this functionality isn't required anymore.

File: zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java [hunk 15]
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Check if the provided {@link Session} is expired.
+     *
+     * @param session the {@link Session} to check.
+     * @return true if expired, false otherwise.
+     */
+    @Override
+    public boolean expired(Session session) {
+        throw new UnsupportedOperationException();
```
daiLlew, line 107: Ditto

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Get the expiry date of the provided {@link Session}
+     *
+     * @param session the  {@link Session} to use.
+     * @return the sessions expiration date time as a {@link Date} instance.
+     */
+    @Override
+    public Date getExpiryDate(Session session) {
```
daiLlew, line 117: ditto

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        throw new UnsupportedOperationException();
+    }
+
+    private Session createZebedeeSession(com.session.service.Session clientSession) {
+        Session session = new Session();
+
+        if (clientSession.getId() == null || StringUtils.isEmpty(clientSession.getId())) {
```
daiLlew, line 124: `id == null` is covered by the `StringUtils.isEmpty` so you can remove this extra check.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+            throw new SessionClientException("client has returned a session with a null/empty id");
+        }
+
+        if (clientSession.getEmail() == null || StringUtils.isEmpty(clientSession.getEmail())) {
```
daiLlew, line 128: ditto

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+     * @param session the  {@link Session} to use.
+     * @return the sessions expiration date time as a {@link Date} instance.
+     */
+    @Override
+    public Date getExpiryDate(Session session) {
+        throw new UnsupportedOperationException();
+    }
+
+    private Session createZebedeeSession(com.session.service.Session clientSession) {
+        Session session = new Session();
```
daiLlew, line 122: **Super nit picky point** but there is no point in creating this object until the validation checks below have passed.  Move it to after the 2 validation checks.

If it fails its wasted effort and memory usage. Granted this is really minuscule saving but stuff like this adds up easily and contributes to a unnecessarily larger memory footprint.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/ZebedeeTestBaseFixture.java
daiLlew: That test case that is failing is passing in a different ID value so the mock case is not matching. I didn't notice this was the test base class so it probably shouldn't be too specific so you can leave it as it is.

daiLlew (changes requested):

daiLlew (approved):

PR #372, opened 2021-01-15
Feature/master keyring

daiLlew:
### Master keyring. 

Initial changes in building/migrating to a master keyring for the collection encryption keys.

- Added new `CollectionKeyStore` implementation and tests.

#### Brief overview
The `CollectionKeyStore` is the class responsible reading/writing & deleting the key files. When a key is written to the store the `key` object is is encrypted with the master keyring encryption key and written to a file on disk -> `$collection_id.key`.

When a key is requested the file is read & decrypted and used to initialise a `SecretKey` object which can be used to decrypt the content of the collection it belongs to.  Permissions and auth for accessing keys will be handled in layer around the keyring, this class is only concerned with persisting/retrieving the keys securely.

### How to review

The code is not currently wired up so code review and/or running the unit tests.

### Who can review
Anyone is welcome to review the code but input from @franmoore05 @janderson2 @CarlHembrough would be welcome.


ci/scripts/audit.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash -eux
 
 pushd zebedee
+    mvn clean install dependency:copy-dependencies -Dmaven.test.skip=true -Dossindex.skip=true
```
daiLlew, line 4: @janderson2 this is a complete hack and really needs to be fixed properly but the audit build passes now for the first time at least 🚀

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeyReadWriterImpl.java
```diff
@@ -0,0 +1,147 @@
...
+     * @param masterIv   the {@link IvParameterSpec} to use when initializing the encryption {@link Cipher}.
+     */
+    public CollectionKeyReadWriterImpl(final Path keyringDir, final SecretKey masterKey, final IvParameterSpec masterIv) {
+        this.keyringDir = keyringDir;
+        this.masterKey = masterKey;
+        this.masterIv = masterIv;
+        this.gson = new GsonBuilder()
+                .registerTypeAdapter(CollectionKey.class, new CollectionKeySerializer())
+                .create();
+    }
```
daiLlew, line 69: **General question:** I am not sure if I need to use a `FileLock` when reading/writing the key files. Keys are individual files and are only ever created, read or deleted never modified I am not sure the lock is necessary. The intention is to store all keys in an in memory cache which will further prevent race conditions to the file.

With this in mind (for now at least) I am planning to leave out the File locking. If it becomes necessary is can be added retrospectively with a reasonably small impact.

ci/scripts/audit.sh
janderson2: @daiLlew I have opened a PR against this branch that I feel might tidy this up a bit but still fixes the build if you don't mind taking a look #373 .

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeyReadWriter.java
```diff
@@ -0,0 +1,27 @@
+package com.github.onsdigital.zebedee.keyring.io;
+
+import com.github.onsdigital.zebedee.keyring.CollectionKey;
+import com.github.onsdigital.zebedee.keyring.KeyringException;
+
+/**
+ * A {@link CollectionKey} read writer is used to read and write Collection keys to/from encrypted files on disk.
+ */
+public interface CollectionKeyReadWriter {
```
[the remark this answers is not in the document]
daiLlew: We do, I will be putting that in as part of another changeset.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeySerializer.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        String secretKeyStr = element.getAsString();
+        if (StringUtils.isEmpty(secretKeyStr)) {
+            throw new JsonParseException(format(FIELD_EMPTY_ERR, SECRET_KEY_FIELD));
+        }
+
+        byte[] keyBytes = Base64.getDecoder().decode(secretKeyStr);
+        SecretKey key = new SecretKeySpec(keyBytes, 0, keyBytes.length, ENCRYPTION_ALGORITHM);
+
```
s-xton, line 58: should zero the keyBytes array after use (have assumed it contains a raw private key).

franmoore05 (commented):
Happy to approve once the comments left by James and Steve have been addressed.   Unit tests all passed

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeySerializer.java
daiLlew: `keyBytes` is defined within the scope of this method so the variable will be discarded after the method returns so I think it's probably unnecessary.
That said it will technically still be on the heap until its garbage collected. I am not sure how much of a risk this is? I am happy to blitz the array if we have any concern this is a potential risk.

s-xton: Useful discussion here: https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords
Will see if there are other viewpoints.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeyReadWriterImpl.java
daiLlew: I have added the file lock in the latest version (pushing soon). It will acquire an exclusive lock on the file so any requests attempting to access it will block until the lock is released.

This will impact performance but it is a necessary hit. However, I'm pretty confident the amount of concurrent activity on each key file will be low so I am hopefully it should be an unnoticeable performance hit.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/CollectionKey.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @return the collection ID the key belongs to.
+     */
+    public String getCollectionID() {
+        return this.collectionID;
+    }
+
+    /**
+     * @return the secret key the used to encrypt/decrypt the collection content.
```
mike-kelly-methods, line 36: Remove extra `the`

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeySerializer.java
daiLlew: This is a very interesting point and that link you have shared is pretty useful & insightful.

However, this presents a bit of problem - I am not sure if it is possible to read or write a `SecretKey` to a json file without creating a string object to hold it. Googling around the approach to storing a []byte in json is to base64 encode it and write it as a String (which is what I have done). However for the reasons covered in that thread it is potentially a security issue as it means the key will have been stored in a String object at some point and its therefore potentially accessible.

I don't want to say it's not possible but Im not sure how to get around it? Open to ideas & or a chat if anyone wants to discuss in detail.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeyStoreImpl.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        Path keyPath = Paths.get(getKeyPath(collectionID));
+
+        if (Files.notExists(keyPath)) {
+            throw new KeyringException(COLLECTION_KEY_NOT_FOUND_ERR, collectionID);
+        }
+
+        try (
+                RandomAccessFile raf = new RandomAccessFile(getKeyPath(collectionID), "rwd");
```
CarlHembrough, line 97: This can be opened as read only?
```suggestion
                RandomAccessFile raf = new RandomAccessFile(getKeyPath(collectionID), "r");
```

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeySerializerTest.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.mockito.Mockito.when;
+
+public class CollectionKeySerializerTest {
+
+    static final String TEST_COLLECTION_ID = "666";
+
+    private CollectionKeySerializer serializer;
+
+    @Mock
+    private JsonElement jsonElement, innerJsonElement;
```
CarlHembrough, line 36: remove `innerJsonElement` if not used

CarlHembrough (changes requested):
A few minor things to consider

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/io/CollectionKeyStoreImpl.java
daiLlew: The access mode is for the lock - as I understand it `rw` mode is an exclusive lock so no one can read the file while this lock has been acquired - which for this is the functionality we need. Whenever we are reading a key file we want to ensure no one else is reading or deleting it.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/store/CollectionKeyStoreImpl.java
```diff
@@ -0,0 +1,193 @@
...
+        try (
+                RandomAccessFile raf = new RandomAccessFile(getKeyPath(collectionID), FILE_ACCESS_MODE);
+                FileChannel channel = raf.getChannel();
+                InputStream in = Channels.newInputStream(channel);
+                CipherInputStream cin = new CipherInputStream(in, getDecryptCipher());
+        ) {
+            // FileLock documentation states the lock will released when the FileChannel used to acquire it is closed.
+            // FileChannel implements autoclosable and in this case is defined inside a try-with-resources which
+            // guarantees auto closeable resourses are closed.
+            channel.lock();
```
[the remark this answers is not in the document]
daiLlew: hmmm. My understanding after reading that documentation is that a FileLock is actually not suitable mechanism for controlling multi thread access to a file. Good spot.

I was hoping to avoid this but I think the only way implement this functionality with files on disk is to use a lock object on for the entire class. I've had to do this elsewhere to fix race condition bugs, it's not nice and it kills performance but I think that is the cost of data integrity when your "database" is files on disk.

(For the record I'm not responsible for that design decision - it's been a long running bone of contention for me).

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/store/CollectionKeyStoreImplTest.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+    }
+
+    @Test(expected = KeyringException.class)
+    public void read_collectionIDEmpty_shouldThrowException() throws Exception {
+        store = new CollectionKeyStoreImpl(keyringDir.toPath(), null, null);
+
+        SecretKey key = null;
+        try {
+            key = store.read("");
```
CarlHembrough, line 71: For the tests that assert on exceptions, the return value is never assigned so is always null. The assertion for `key` could be removed. (same applies for other tests)

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/store/CollectionKeyStoreImplTest.java
```diff
@@ -0,0 +1,269 @@
...
+    IvParameterSpec createNewInitVector() {
+        byte[] iv = new byte[16];
+
+        SecureRandom random = new SecureRandom();
+        random.nextBytes(iv);
+
+        return new IvParameterSpec(iv);
+    }
+
+    byte[] encyrpt(String plainText, SecretKey key, IvParameterSpec iv) throws Exception {
```
CarlHembrough, line 250: typo 
```suggestion
    byte[] encrypt(String plainText, SecretKey key, IvParameterSpec iv) throws Exception {
```

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/store/CollectionKeyStoreImplTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    byte[] decrypt(byte[] encryptedMessage, SecretKey key, IvParameterSpec iv) throws Exception {
+        Cipher cipher = Cipher.getInstance(CIPHER_ALGORITHM);
+        cipher.init(Cipher.DECRYPT_MODE, key, iv);
+
+        byte[] bytes = Base64.getDecoder().decode(encryptedMessage);
+        return cipher.doFinal(bytes);
+    }
+
+    void createPlainTextCollectionKeyFile(String collectionID, SecretKey secretKey) throws Exception {
```
CarlHembrough, line 265: `secretKey` is unused

CarlHembrough (changes requested):
Much improved - a few minor things to check out

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/store/CollectionKeyStoreImplTest.java
daiLlew: What the hell is wrong with my typing?

CarlHembrough (approved):
LGTM 👍

PR #374, opened 2021-01-19
Feature/keyring impl

daiLlew:
### What

- Added new `Keyring` object with implementation of `add` method and unit tests.
- Updated Methods in `CollectionKeyStore` to `synchronised` as this is a more appropriate mechanism for ensuring the object is thread safe.

The keyring is an object used to hold collection encryption keys in a central place. When a user needs to view collection content the key will be retrieved from this central keyring and used to decrypt the content on read and encrypt it on write.

Keyring uses `CollectionKeyStore` to read/write entries to/from storage but also holds the key data in an in-memory cache for speed. 

### Important points to consider:
- The store and the cache should be kept in sync (when an entry is added it should be added to both the cache and the store).
- The new keyring is used as a central place to store the collection keys so it needs to be thread safe and maintain data integrity when accessed concurrently.

### How to review

Code is not currently wired up so will have no impact on live code. Code review and unit tests will be sufficient.

### Who can review

Any one.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/KeyringImpl.java
```diff
@@ -0,0 +1,137 @@
...
+
+import com.github.onsdigital.zebedee.keyring.store.CollectionKeyStore;
+import liquibase.util.StringUtils;
+
+import javax.crypto.SecretKey;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * In memory {@link Keyring} implmentation. Keyring uses a {@link CollectionKeyStore} to persist entries to storage
```
justinpjose, line 11:
```suggestion
 * In memory {@link Keyring} implementation. Keyring uses a {@link CollectionKeyStore} to persist entries to storage

File: zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/KeyringImpl.java [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+ * whilst maintaining a copy of the data in an in-memory cache for speedy retreival. If an attempt is made to add a
```
justinpjose, line 12:
```suggestion
 * whilst maintaining a copy of the data in an in-memory cache for speedy retrieval. If an attempt is made to add a
```

justinpjose (commented):
Just some small spelling mistakes. But the code is awesome overall - very neat and organized

justinpjose (commented):
The unit test passes and just remember that

franmoore05 (approved):

PR #380, opened 2021-01-27
Update imports from package name change and added validation

morses-code:
### What

Updated to newer version of `dp-sessions-service-java-client`
Updated imports
Added validation to `get()` to check for empty session identifiers

### How to review

Run publishing stack with `dp-sessions-api` and local instance of redis (docker)

`docker run -p 6379:6379 --name redis -d redis:6.0 redis-server --requirepass "default"`

Note -- may have issue with ports as `dp-frontend-homepage-controller` has default config conflicting with `dp-sessions-api`

### Who can review

Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
@@ -3,15 +3,13 @@ package com.github.onsdigital.zebedee.session.service;
 import com.github.onsdigital.zebedee.reader.util.RequestUtils;
 import com.github.onsdigital.zebedee.session.model.Session;
 import com.github.onsdigital.zebedee.user.model.User;
-import com.session.service.client.SessionClient;
-import com.session.service.client.SessionClientImpl;
-import com.session.service.entities.SessionCreated;
-import com.session.service.error.SessionClientException;
+import com.github.onsdigital.session.service.client.SessionClient;
+import com.github.onsdigital.session.service.entities.SessionCreated;
+import com.github.onsdigital.session.service.error.SessionClientException;
 import org.apache.commons.lang.StringUtils;
 
 import javax.servlet.http.HttpServletRequest;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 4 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -73,67 +72,65 @@ public Session get(HttpServletRequest id) throws IOException {
...
      * @return a {@link Session} for the requested email if it exists and is not expired. Return null otherwise.
      * @throws IOException for any problems getting the session.
      */
     @Override
-    public Session find(String email) throws IOException {
-        com.session.service.Session clientSession = client.getSessionByEmail(email);
-
-        Session session = createZebedeeSession(clientSession);
-        return session;
+    public Session find(String email) throws IOException, SessionClientException {
```
daiLlew, line 91: Do the same check here

`        if (StringUtils.isEmpty(email)) {
            return null;
        }`

Can  you also update the javadoc to state it will return null if the email is null or empty please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/NewSessionsServiceImpl.java
```diff
@@ -68,72 +67,70 @@ public Session get(HttpServletRequest id) throws IOException {
...
      * @param id the ID of the session to get,
-     * @return the {@link Session} instance if it exists and is not expired.
+     * @return the {@link Session} instance if it exists and is not expired. If the id is null or empty will return null.
      * @throws IOException for any problems getting the session.
      */
     @Override
     public Session get(String id) throws IOException {
-        com.session.service.Session clientSession = client.getSessionByID(id);
+        if (StringUtils.isEmpty(id)) {
+            return null;
```
CarlHembrough, line 76: Consider throwing an exception here instead of returning null. That way you can add detail to the exception message telling the caller exactly what the issue is. We should probably add test coverage for these cases as well.

morses-code: Hi Carl. Thanks for the comment. Their is a reason this needs to return a null.
When adding a collection the keychain checks anyone with access to the collection. If that user hasn't logged in it has no session so returns an error when it doesn't really need to. Not sure if there is a bit more graceful of a way around it at the minute.

daiLlew: @CarlHembrough we did this initially but we broke backwards compatibility. The current / existing functionality expects this to be `null` in some cases. The ripple effect of doing this "properly" is pretty wide spread so we are going to keep this as null.

daiLlew (approved):

PR #387, opened 2021-02-05
Send slack notifications on startup for collection approval status IN_PROGESS or ERROR

CarlHembrough:
### What
Send slack notifications on startup for collection approval status IN_PROGESS or ERROR
  - remove redundant indexPublishedCollections function.
  - move the loading of collections out of the `loadExistingCollectionsIntoScheduler` function, allowing it to be reused in the `alertOnInProgressCollections` function.
  - Added Notifier interface and SlackNotifier implementation
  - Add @deprecated annotation to the existing SlackNotification class

### How to review
Review changes / tests. If you want to test locally you can create a collection in Florence and manually change the approval state in the collection json file.

### Who can review
Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
@@ -8,7 +8,6 @@ import com.github.onsdigital.zebedee.configuration.Configuration;
 import com.github.onsdigital.zebedee.exceptions.BadRequestException;
 import com.github.onsdigital.zebedee.exceptions.NotFoundException;
 import com.github.onsdigital.zebedee.exceptions.UnauthorizedException;
-import com.github.onsdigital.zebedee.json.ApprovalStatus;
 import com.github.onsdigital.zebedee.json.serialiser.IsoDateSerializer;
 import com.github.onsdigital.zebedee.model.Collection;
 import com.github.onsdigital.zebedee.model.Collections;
@@ -19,7 +18,6 @@ import com.github.onsdigital.zebedee.model.publishing.scheduled.PublishScheduler
 import com.github.onsdigital.zebedee.model.publishing.scheduled.Scheduler;
 import com.github.onsdigital.zebedee.reader.configuration.ReaderConfiguration;
 import com.github.onsdigital.zebedee.user.model.User;
-import com.github.onsdigital.zebedee.util.SlackNotification;
 import com.github.onsdigital.zebedee.util.slack.Notifier;
 import com.github.onsdigital.zebedee.util.slack.SlackNotifier;
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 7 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
```diff
@@ -122,9 +125,16 @@ public static void init() {
...
         SlackNotification.alarm("Zebedee has just started. Ensure an administrator has logged in.");
 
-        loadExistingCollectionsIntoScheduler();
+
+        try {
+            Collections.CollectionList collections = zebedee.getCollections().list();
+            alertOnInProgressCollections(collections, new SlackNotifier());
+            loadExistingCollectionsIntoScheduler(collections);
+        } catch (IOException e) {
+            error().logException(e, "zebedee root: failed to load collections list check in progress approvals");
```
jessjenkins, line 134: My only thought is if the loading fails for some other reason other than ‘in progress’ ones, would the error message be misleading?

CarlHembrough: Good spot Jess - I have updated the log message so that it's not specific to any one function call that uses the collection list.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/Notifier.java
```diff
@@ -0,0 +1,17 @@
+package com.github.onsdigital.zebedee.util.slack;
+
+import com.github.onsdigital.zebedee.model.Collection;
+
+
+/**
+ * Notifier is a generic interface for sending notifications to external systems such as Slack
+ */
+public interface Notifier {
```
daiLlew, line 9: I think the SlackNotification util is used a number of times in the Root class, possibly scope creep but it feels like if we are defining an interface to wrap around this util we should probably include all the methods being used in the Root class. For consistency if nothing else.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/Notifier.java
```diff
@@ -0,0 +1,17 @@
...
+
+    /**
```
daiLlew, line 11: Missing description in Javadoc. You've explained the params but not when the method does 😂

daiLlew (changes requested):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
```diff
@@ -194,6 +187,18 @@ private static void loadExistingCollectionsIntoScheduler() {
...
+    static void alertOnInProgressCollections(Collections.CollectionList collections, Notifier notifier) {
+        info().log("zebedee root: checking existing collections for in progress approvals");
+        collections.stream()
+                .filter(c -> c.getDescription().approvalStatus == ApprovalStatus.IN_PROGRESS
+                        || c.getDescription().approvalStatus == ApprovalStatus.ERROR)
+                .forEach(c -> {
+                    info().data("collectionId", c.getDescription().getId())
+                            .data("type", c.getDescription().getType().name())
+                            .log("zebedee root: collection approval is in error or in progress state on zebedee startup");
+                    notifier.collectionAlarm(c, "Collection approval is in IN_PROGRESS or ERROR state on zebedee startup. It may need to be re-approved manually.");
```
daiLlew, line 199: You could probably make the message a const as its the same message each time. Plus keeps like line short 😁

CarlHembrough: I debated doing this, but decided not to as the slack message has additional formatting and the suggested fix. As I understand the log message should be all lower case without formatting.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -991,6 +990,12 @@ public boolean hasCollection(String name) {
             return getCollectionByName(name) != null;
         }
 
+        public List<Collection> withApprovalInProgress() {
```
daiLlew, line 993: Technically - `withApprovalInProgressOrError`

daiLlew (approved):

PR #392, opened 2021-02-09
added new secureKeyring impl + unit tests

daiLlew:
### What

Created new `SecureKeyring` implementation and added unit tests for new functionality.
- Added `populateFromUser` method.
- Added unit tests for new functionality.

### How to review

The intention is for `SecureKeyring` to be a permissions wrapper around the new `Keyring` implementation. This keeps the keyring & permissions logic separate allowing us to transition over to the new world Permissions API without any changes required to the core keyring logic.

`populateFromUser` populates the new central keyring from the legacy keyring of an authenticated `User`. This is a temporary hook to allows us to migrate from the old keyring to the new keyring without losing any of the existing collection keys. The plan is to "dual run" the legacy and new keyring implementations for a bit. Once we're confident its all working as expected we can remove the legacy keyring without losing any data.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/SecureKeyring.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/SecureKeyring.java
@@ -7,7 +7,7 @@ public interface SecureKeyring {
     /**
      * Populate the Keyring from an unlocked {@link User#keyring()}
      *
-     * @param user the user to use to populate the keyring.
+     * @param user the user that populates the keyring.
      * @throws KeyringException problem populating the keyring.
      */
     void populateFromUser(User user) throws KeyringException;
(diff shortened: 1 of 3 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/SecureKeyring.java
```diff
@@ -0,0 +1,14 @@
+package com.github.onsdigital.zebedee.keyring;
+
+import com.github.onsdigital.zebedee.user.model.User;
+
+public interface SecureKeyring {
+
+    /**
+     * Populate the Keyring from an unlocked {@link User#keyring()}
+     *
+     * @param user the user to use to populate the keyring.
```
morses-code, line 10: Reading this back feels a bit too wordy.... maybe this would read better:
`@param user that populates the keyring` ?

morses-code (commented):
One comment about the java doc.

morses-code (approved):

PR #396, opened 2021-02-11
Integrate Image API into collection publishing

jessjenkins:
### What

Add the image API client into zebedee so that images can be published at the same time as a collection.
There is a new feature flag (`ENABLE_IMAGE_PUBLISHING`) which defaults to off, however a secrets PR will follow to manually set it so as well as to configure the `IMAGE_API_URL` for dev and prod environments.

### How to review

Check code and tests pass etc.
I have tested locally with manual and scheduled collections as well as with the feature disabled, but if you wish to do so you will need an imported image with a collection ID matching a publishable collection (this will be done automatically by the homepage functionality in florence when it's all integrated but for now you can import one with `make produce` under the dp-image-importer and then manually change the collection id in the image using a mongo client.

### Who can review

Anyone but me (@CarlHembrough has already helped in the development)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -172,8 +175,16 @@ public static boolean publish(Collection collection, String email, CollectionRea
                         return false;
                     }
 
+                    Future<Boolean> imageFuture = publishImages(collection);
+
                     publishComplete = publishFilesToWebsite(collection, email, collectionReader);
 
+                    try {
+                        publishComplete &= imageFuture.get().booleanValue();
```
CarlHembrough, line 183: Has it been confirmed that failing to publish images will result in the publish being deemed a failure? If I am reading the code correctly, updated pages could be published to the website, but if the images fail then the publish will be labelled as failed, even though the pages have published.

CarlHembrough, line 183: Consider using `&&` instead of `&=`. As I understand using `&&` will not attempt to evaluate the right hand side if publishComplete is already false. i.e. if the files fail to publish, it won't bother waiting to see if the images publish.

https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.23

I am thinking of a scenario where a publish fails, and for some reason if the image publishing hangs.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -692,6 +703,25 @@ private static boolean publishDatasets(Collection collection) throws IOException
         return datasetsPublished;
     }
 
+    private static Future<Boolean> publishImages(Collection collection){
+        String collectionId = collection.getDescription().getId();
+
+        return pool.submit(() -> {
```
CarlHembrough, line 709: Consider having a separate thread pool for publishing images / anything that isn't traditional publishing. If for some reason the images publish is failing and hanging, it could end up hogging all of the threads in the pool, preventing other publishes from taking place.

Some more explanation of the idea here https://stackoverflow.com/questions/26243422/is-having-a-single-threadpool-better-design-than-multiple-threadpools

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -692,6 +703,25 @@ private static boolean publishDatasets(Collection collection) throws IOException
...
+            Boolean complete = false;
+            try {
+                imageServiceSupplier.getService().publishImagesInCollection(collection);
+                complete = true;
+            } catch (Exception e) {
+                error().data("collectionId", collectionId).data("publishing", true)
+                        .logException(e, "Exception publishing images via image API");
+
+                PostMessageField msg = new PostMessageField("Error", e.getMessage(), false);
+                collectionAlarm(collection, "Exception publishing images via image API", msg);
```
CarlHembrough, line 719: I tried to avoid any error reporting within the core publishing code when it was originally put in. I guess there would be two schools of thought here:

1 - prevent any error reporting in the core publish code. Capture any exception and report it later once we're outside of the publish. Imagine a scenario where slack is down and the call to slack hangs. It would hold up the publish process. (This is where using `&&` in the point above may be useful)
2 - ensure that the image publishing code is sufficiently isolated from the old publish code, so that if it fails it has minimal impact on the existing publish. (This is where an isolated thread pool would be preferred)

CarlHembrough (changes requested):
Changes are looking good generally. I have added some points for consideration

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -119,6 +128,14 @@ public static boolean executePublish(Collection collection, CollectionReader col
             success &= publishDatasets(collection);
         }
 
+        if (imageFuture != null) {
+            try {
+                success = success && imageFuture.get();
+            } catch (InterruptedException | ExecutionException e) {
+                success = false;
```
daiLlew, line 135: I can't remember what was discussed and agreed on what happens to the publish if this fails . I think we said the publish should carry on regardless.

I think we should be logging out these errors for visibility on what went wrong.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
daiLlew: I agree with Carl's point here. 

I think the suggested way to handle exceptions thrown in a thread executor service  is to just let it be thrown. Once the tasks have completed you iterate over the list of futures calling  `future.get()` on each one. If a task failed `get` will throw an exception. You then catch this exception and handle it as appropriate.

You're error handling logic is spot on I think you just need to move entire catch block move it up to the catch block on line `135` where you are iterating over the futures.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeImageService.java
```diff
@@ -0,0 +1,41 @@
...
+import com.github.onsdigital.dp.image.api.client.model.Image;
+import com.github.onsdigital.dp.image.api.client.model.Images;
+import com.github.onsdigital.zebedee.model.Collection;
+
+import java.io.IOException;
+
+/**
+ * Image related services
+ */
+public class ZebedeeImageService implements ImageService {
```
daiLlew, line 14: No need to call it `ZebedeeImageService` - we're in Zebedee so that is a given. `ImageServiceImpl` would follow the naming convention for most interface implementations in this codebase.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeImageService.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private ImageClient imageClient;
+
+    public ZebedeeImageService(ImageClient imageClient) {
```
daiLlew, line 18: Javadoc.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -119,6 +128,14 @@ public static boolean executePublish(Collection collection, CollectionReader col
             success &= publishDatasets(collection);
         }
 
+        if (imageFuture != null) {
```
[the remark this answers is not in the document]
jessjenkins: This is how I had it initially, but then the linting was moaning because I was not checking for a null before doing `imageFuture.get()`. If I put it under a feature flag instead, should I just ignore the null (im)possibility or have an exception to throw that will never occur?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ZebedeeImageService.java
jessjenkins: I was keeping it consistent with the existing `ZebedeeDatasetService` which it is based on but I agree it is a better name. I'm renaming it to your suggestion.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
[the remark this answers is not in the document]
jessjenkins: I have updated the code to use the feature flag and throw an exception on nulls instead.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
jessjenkins: I have moved the catch block from the image future here so the logs for all exceptions are now handled consistently.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
jessjenkins: I have moved the catch block out of this future and to the code that completes the future in the calling method.

CarlHembrough (approved):
LGTM - good work 👍

PR #399, opened 2021-02-16
Remove deprecated methods from the Sessions interface

CarlHembrough:
### What
Remove deprecated methods from the Sessions interface

- Add implementation for SessionsAPIServiceImpl.expired() function, to prevent exception being thrown
- The methods are only used internally by the SessionsServiceImpl, so they have been removed from the public interface. They have been made private / package private in the SessionsServiceImpl.
- Remove SessionsServiceImpl.exists() function as it is only used in unit tests.
- remove unused usersService in SessionsServiceServiceImplTest.java
- remove redundant unit tests / mocking / assertions

### How to review
Review changes / check tests

### Who can review
Anyone


daiLlew (dismissed):
I think this should work.

daiLlew (approved):

PR #408, opened 2021-03-04
Skip publishing of created/deleted images

jessjenkins:
### What

When publishing images in a collection, skip over any that have been deleted or that have been created but not uploaded. The latter will be common because image records are created as soon as the user opens the homepage section in florence and if they cancel without uploading the record will be forgotten. We should not try to publish these images it will cause an exception that will fail the publish of that collection. 

Also improve logging of image publishing.

### How to review

Check the code and ensure a test is in place to cover this scenario.

### Who can review

Anyone but me

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServiceImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServiceImpl.java
@@ -8,11 +8,15 @@ import com.github.onsdigital.zebedee.model.Collection;
 
 import java.io.IOException;
 
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
+
 /**
  * Image related services
  */
 public class ImageServiceImpl implements ImageService {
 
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServiceImpl.java
```diff
@@ -32,9 +36,25 @@ public ImageServiceImpl(ImageClient imageClient) {
...
+                            .data("imageId", image.getId())
+                            .data("imageState",image.getState())
+                            .log("skipping publish of image");
+                    break;
+                default:
+                    info().data("publishing", true).data("collectionId", collection.getId())
+                            .data("imageId", image.getId())
+                            .log("publishing image");
+                    imageClient.publishImage(image.getId());
+            }
```
morses-code, line 57: Maybe just a preference of mine maybe, but could this be refactored to its own private method?
Happy for this to be ignored and happy to approve otherwise.

morses-code (approved):

PR #412, opened 2021-03-09
Feature/legacy keyring

daiLlew:
### Legacy Keyring.

Created _new_ legacy keyring implementation. This class duplicated the existing keyring functionality putting behind the newly created `Keyring` interface. 

Once complete we will update all uses of the keyring to go via the interface. This allows us to switch implementations without breaking the code.


franmoore05 (approved):
Changes look good to me

PR #423, opened 2021-03-18
Slack alert then ignore unpublished images

jessjenkins:
### What

When publishing a collection containing uploaded images…
- Skip publishing for any images in an `importing` or `failed-import` state
- Send a slack alert so the development team can look into why they failed

### How to review

Check the code and ensure there are appropriate tests (where possible, it is zebedee after all)
Ensure all tests pass

NB. If you want to try running it, you'll need to setup your local zebedee to send slack messages. For example, I have my own dev slack organisation and I added the following to my run-cms.sh
```
export slack_alarm_channel="#zeb-alarm"
export slack_default_channel="#zeb-system"
export slack_publish_channel="#zeb-publog"
export slack_api_token="<my slack api token>"
```

You'll then have to have a collection to publish that has at least one image in one of the states above.

### Who can review

Anyone but me

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
@@ -20,11 +20,13 @@ import com.github.onsdigital.zebedee.reader.CollectionReader;
 import com.github.onsdigital.zebedee.reader.Resource;
 import com.github.onsdigital.zebedee.service.DatasetService;
 import com.github.onsdigital.zebedee.service.ImageService;
+import com.github.onsdigital.zebedee.service.ImageServicePublishingResult;
 import com.github.onsdigital.zebedee.service.ServiceSupplier;
 import com.github.onsdigital.zebedee.util.Http;
 import com.github.onsdigital.zebedee.util.SlackNotification;
 import com.github.onsdigital.zebedee.util.ZebedeeCmsService;
 import com.github.onsdigital.zebedee.util.slack.PostMessageField;
+import com.github.onsdigital.zebedee.util.slack.SlackNotifier;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.IOException;
@@ -106,7 +108,7 @@ public class Publisher {
             throws IOException {
         boolean success = false;
 
-        Future<Boolean> imageFuture = null;
+        Future<ImageServicePublishingResult> imageFuture = null;
         if (CMSFeatureFlags.cmsFeatureFlags().isImagePublishingEnabled()) {
             imageFuture = publishImages(collection);
         }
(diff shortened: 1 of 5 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServicePublishingResult.java
```diff
@@ -0,0 +1,44 @@
+package com.github.onsdigital.zebedee.service;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ImageServicePublishingResult {
+    private int totalImages;
+    private List<ImageDetails> unpublishedImages = new ArrayList<>();
+
+    public ImageServicePublishingResult(int totalImages) {
```
daiLlew, line 10: Do we care if this is a negative number? What would happen to the code using this data if it returned a negative value? Possibly an unlikely scenario but I thought I'd ask the question all the same.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServiceImpl.java
```diff
@@ -33,40 +35,64 @@ public ImageServiceImpl(ImageClient imageClient) {
...
         info().data("publishing", true).data("collectionId", collection.getId())
                 .data("numberOfImages", images.getTotalCount())
                 .log("publishing images in collection");
 
+        ImageServicePublishingResult result = new ImageServicePublishingResult(images.getTotalCount());
         for (Image image : images.getItems()) {
-            publishImage(image, collection.getId());
+            PublishStatus status = publishImage(image, collection.getId());
+            if (PublishStatus.UNPUBLISHED.equals(status)) {
+                result.addUnpublishedImage(image.getId(), image.getState());
```
daiLlew, line 49: Is there any kind of "reason" type data that could be included in this? Just wondering if its available and if it would be useful. Possibly out of scope but asking the question.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServiceImpl.java
```diff
@@ -33,40 +35,64 @@ public ImageServiceImpl(ImageClient imageClient) {
...
         }
+
+        return result;
     }
 
     /**
      * Publish an image unless its state is 'created' or 'cancelled'
      */
-    private void publishImage(Image image, String collectionId) throws IOException, ImageAPIException {
+    private PublishStatus publishImage(Image image, String collectionId) throws IOException, ImageAPIException {
```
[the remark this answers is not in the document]
jessjenkins: Yes, in the `default` block, for images that aren't skipped, it publishes the image.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServiceImpl.java
jessjenkins: I was keeping it as small as possible. This is purely for use in reporting to slack so as little info as possible to keep the message small. We could always expand another time if it's deemed necessary.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServicePublishingResult.java
```diff
@@ -0,0 +1,44 @@
...
+    public int getTotalImages() {
+        return totalImages;
+    }
+
+    public List<ImageDetails> getUnpublishedImages() {
+        return unpublishedImages;
+    }
+
+    public void addUnpublishedImage(String id, String status) {
+        unpublishedImages.add(new ImageDetails(id, status));
```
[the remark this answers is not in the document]
jessjenkins: Ummmm, yes AND no? Yes, if I'm writing properly robust Java then of course, but given we're adding a simple function to already complicated code in a deprecated service I went down the 'YAGNI' route knowing the only thing calling this would have values. I should probably add Exceptiony gubbins but given how sensitive the exception handling in this area is I was nervous. Happy to take advice.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/ImageServicePublishingResult.java
```diff
@@ -0,0 +1,44 @@
+package com.github.onsdigital.zebedee.service;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ImageServicePublishingResult {
```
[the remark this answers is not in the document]
jessjenkins: Added 👍

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -133,7 +135,11 @@ public static boolean executePublish(Collection collection, CollectionReader col
                 if (imageFuture == null) {
                     throw new Exception("image future unexpectedly null on completion of publishing");
                 }
-                success = success && imageFuture.get();
+                ImageServicePublishingResult result = imageFuture.get();
+
+                if (result.getUnpublishedImages() != null && result.getUnpublishedImages().size() > 0) {
```
daiLlew, line 140: Sorry I missed this before - but we should check `result != null` too. Otherwise `result.getUnpublishedImages()` will throw a null pointer ex.

daiLlew (approved):

PR #429, opened 2021-03-24
refactored add and tests, split out tests as they were getting too bi…

daiLlew:
### What

Refactor `add` - previously add updated a single user, however in order to migrate to the new keyring we need to move the "distribute key to users logic" behind the `Keyring` interface.  I have updated the add method to include this functionality - essentially it now gives the key to all users who should have it and removes it from any without permission.

I have also updated the tests 😖  to prove this functionality.

**Note**: This is a complex class that has multiple responsibilities which is normally not good but 
- This is unavoidable it's how the original implementation was written.
- This is temp code which will be deleted once we switch over 🎉 

The downside of this is the tests scenarios kept growing exponentially. To keep my sanity and to make it easier to read & maintain to I have split out the test cases for each method in the class into their own test class.

### How to review

Code review.

### Who can review

Anyone.


nshumoogum (changes requested):
Looks mostly fine to me, just 1 typo spotted. Unit tests pass  but may need someone who understands the business logic/criteria to review as well.

nshumoogum (approved):

PR #438, opened 2021-03-29
Remove InfluxDB as it is no longer used

CarlHembrough:
### What
This task was originally to fix a vulnerable dependency (okHTTP). I seen the the dependency was only used by the InfluxDB class, which is no longer used. Also removed some other unused legacy things along the way.

- Remove unused legacy InfluxDB class
- Remove InfluxDB dependency
- Remove related configuration
- Remove unused legacy PublishTask class
- Remove cli.sh - it refers to the zebedee-cli component which no longer exists

### How to review
Sanity check

### Who can review
Anyone


daiLlew (approved):
Assuming the build passes 👍

daiLlew (approved):

PR #442, opened 2021-03-31
Feature/collections api keyring

daiLlew:
### What

Refactor `Collection` GET endpoint to make it testable in preparation for Keyring update.

I've also refactored `CollectionDescription` making the fields private and added `getter` & `setter` methods to follow Java best practice & standards. The rest of the changes are updated to all the places using `CollectionDescription` - they now have to use the getter/setter methods instead of accessing fields directly.

### How to review

Review code and check unit tests.

### Who can review

Anyone.


jessjenkins (approved):
Soooo much cleaner! 👏

PR #445, opened 2021-04-01
added populate method to new keyting interface

daiLlew:
### What

Support for legacy keyring. When a new user is created their keyring has to be populate for the collections they have access to. The keys to assign as copied from the admin user's (who is creating the new user) keyring.

This functionality existing in  the permissions service currently - moving it behind new interface to allow migration. It will be deleted once we have migrated to the new keyring as its not a problem that exists in that world.

### How to review

CoDe rEvIEw.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
@@ -51,7 +51,7 @@ public class LegacyKeyringImpl implements Keyring {
     static final String SAVE_USER_KEYRING_ERR = "error saving changes to user keyring";
 
     private Sessions sessions;
-    private UsersService users;
+    private UsersService usersService;
     private PermissionsService permissions;
     private KeyringCache cache;
     private ApplicationKeys applicationKeys;
@@ -63,11 +63,11 @@ public class LegacyKeyringImpl implements Keyring {
      * @param cache           the {@link KeyringCache} to use.
      * @param applicationKeys the {@link ApplicationKeys} to use.
      */
-    public LegacyKeyringImpl(final Sessions sessions, final UsersService users, PermissionsService permissions,
+    public LegacyKeyringImpl(final Sessions sessions, final UsersService usersService, PermissionsService permissions,
                              final KeyringCache cache,
                              final ApplicationKeys applicationKeys) {
         this.sessions = sessions;
-        this.users = users;
+        this.usersService = usersService;
         this.permissions = permissions;
         this.cache = cache;
         this.applicationKeys = applicationKeys;
@@ -233,7 +233,7 @@ public class LegacyKeyringImpl implements Keyring {
 
     private UserList listUsers() throws KeyringException {
         try {
-            return users.list();
+            return usersService.list();
         } catch (IOException ex) {
             throw new KeyringException(LIST_USERS_ERR, ex);
         }
@@ -246,7 +246,7 @@ public class LegacyKeyringImpl implements Keyring {
         }
 
         try {
-            users.addKeyToKeyring(user.getEmail(), collection.getDescription().getId(), key);
+            usersService.addKeyToKeyring(user.getEmail(), collection.getDescription().getId(), key);
         } catch (IOException ex) {
             throw new KeyringException(ADD_KEY_SAVE_ERR, ex);
         }
@@ -259,7 +259,7 @@ public class LegacyKeyringImpl implements Keyring {
         }
 
         try {
-            users.removeKeyFromKeyring(user.getEmail(), collection.getDescription().getId());
+            usersService.removeKeyFromKeyring(user.getEmail(), collection.getDescription().getId());
         } catch (IOException ex) {
             throw new KeyringException(REMOVE_KEY_SAVE_ERR, ex);
         }
@@ -335,7 +335,7 @@ public class LegacyKeyringImpl implements Keyring {
             }
         }
 
-        saveKeyring(target);
+        saveKeyringChanges(target);
     }
 
     private void validateUser(User user) throws KeyringException {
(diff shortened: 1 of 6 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
```diff
@@ -306,6 +309,35 @@ public void unlock(User user, String password) throws KeyringException {
...
+            throw new KeyringException(CACHE_KEYRING_NULL_ERR);
+        }
+
+        if (!srcKeyring.isUnlocked()) {
+            throw new KeyringException(KEYRING_LOCKED_ERR);
+        }
+
+        validateUser(target);
+
+        if (collectionIDs == null || collectionIDs.isEmpty()) {
```
sandypadmanabhan, line 327: Do we want to check this before getting the cachedUserKeyring?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl_PopulateTest.java
```diff
@@ -0,0 +1,209 @@
...
+                .thenReturn("");
+
+        KeyringException ex = assertThrows(KeyringException.class,
+                () -> legacyKeyring.populate(srcUser, targeUser, null));
+
+        assertThat(ex.getMessage(), equalTo(EMAIL_EMPTY_ERR));
+    }
+
+    @Test
+    public void testPopulate_targetUserKeyringNull_ShouldThrowException() {
```
sandypadmanabhan, line 160: To begin with the target user Keyring is null should we not return a new HashSet?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
```diff
@@ -361,4 +393,12 @@ private void validatePassword(String password) throws KeyringException {
             throw new KeyringException(PASSWORD_EMPTY_ERR);
         }
     }
+
+    private void saveKeyring(User user) throws KeyringException {
+        try {
+            users.updateKeyring(user);
+        } catch (IOException ex) {
+            throw new KeyringException(SAVE_USER_KEYRING_ERR, ex);
+        }
```
sandypadmanabhan, line 402: target.keyring() has already been updated before the saveKeyring function is called - is this an expected behaviour?

sandypadmanabhan: The IOException can occur after the keyring has been updated to the target user- is this an expected behaviour?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
```diff
@@ -46,6 +46,9 @@
     static final String UNLOCK_KEYRING_ERR = "error unlocking user keyring";
     static final String GET_KEY_RECIPIENTS_ERR = "error getting recipients for collection key";
     static final String LIST_USERS_ERR = "error listing all users";
+    static final String CACHE_KEYRING_NULL_ERR = "expected cached keyring but was not found";
+    static final String KEYRING_LOCKED_ERR = "cached has not been unlocked";
```
sandypadmanabhan, line 50: Should this error say 'cached keyring has not been unlocked'?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
daiLlew: Some context might help explain this a bit more clearly:

When we create a new user their keyring needs to be populate with the keys for the collection they should have access too. In the current world this is done by **reading the key values from the user who created** the new user.

You can write a new key to a keyring at anytime (even if its locked) but you can only read from a keyring that has been unlocked. When a user logs in their keyring is unlocked and added to the cache - this enables us to read keys from it later on without requiring their password each time.

In the context of this functionality if the keyring of the user who is is creating a new user is not in the cache then something has gone wrong and we cannot read any values from it. Even if we don't have any keys to add this is scenario would be the symptom of something having gone very wrong which is why we should check it. This of it as a mandatory prerequisite and it's not met we need to sound the alarm

If everything is as it should be and there are no keys to add then we essentially don't need to do anything. However, I think I could move this check above `validateTargetUser()` as it would be more efficient i.e. only valid this target user if we know we need to add keys.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
daiLlew: > target.keyring() has already been updated before the saveKeyring function is called - is this an expected behaviour?

Yes.  `user.updateKeyring()` takes a user object and saves the changes to the keyring field writing them to disk to persist the changes. The method name isn't very clear but it's existing functionality and don't want to mess with it - scope creep. I will rename `users` though to `userService` to make it a bit clearer.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl_PopulateTest.java
daiLlew: We could but the expectation is this should not be null. If it is null something has gone wrong and we should throw an error to draw attention to this. As you've suggested we could "recover" this situation by simply creating a new map but this would be hiding the fact something was not right/not as expected. 

Others might disagree with me on this by my philosophy is it's almost always better to error when something it not as expected and draw attention to it. 

When that happens we can investigate then and work out if its OK or if there really is an issue and take the appropriate action.
If we hide that something has gone wrong it could cause other problems in different parts of the app or manifest itself in subtle ways which can cause confusion or weird behaviour.

jessjenkins (approved):

PR #446, opened 2021-04-07
refactor scheduler cache to work for new world

daiLlew:
### What

Refactor the existing scheduler cache. I've decoupled it from the legacy Keyring and implemented it behind a new interface and updated all existing uses of it to use the new interface. 

When we migrate to the new keyring we can swap the implementations without breaking any code.

### How to review

Describe the steps required to test the changes.

### Who can review

Describe who worked on the changes, so that other people can review.


sandypadmanabhan (approved):
Looks good to me

PR #457, opened 2021-05-18
Added noop StartUpAlerterImpl as default impl

daiLlew:
### What

- Added a No Op StartUpAlerterImpl to use as default if no config is provided. 
- Refactor StartUpAlerter to log exceptions instead of throwing them. If the alert fails the CMS will now continue to function.
- Refactored StartUpAlerter to be more resilient to error scenarios. Logs errors instead of throwing exceptions to offer degraded functionality over total loss of functionality.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
@@ -47,6 +47,7 @@ import com.github.onsdigital.zebedee.teams.store.TeamsStoreFileSystemImpl;
 import com.github.onsdigital.zebedee.user.service.UsersService;
 import com.github.onsdigital.zebedee.user.service.UsersServiceImpl;
 import com.github.onsdigital.zebedee.user.store.UserStoreFileSystemImpl;
+import com.github.onsdigital.zebedee.util.slack.NoOpStartUpAlerter;
 import com.github.onsdigital.zebedee.util.slack.StartUpAlerter;
 import com.github.onsdigital.zebedee.util.slack.StartUpAlerterImpl;
 import com.github.onsdigital.zebedee.util.versioning.VersionsService;
@@ -59,6 +60,7 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
 
 import static com.github.onsdigital.logging.v2.event.SimpleEvent.error;
 import static com.github.onsdigital.logging.v2.event.SimpleEvent.info;
@@ -229,13 +231,7 @@ public class ZebedeeConfiguration {
 
         imageService = new ImageServiceImpl(imageClient);
 
-        SlackClient slackClient = new SlackClientImpl(new Profile.Builder()
-                .emoji(":flo:")
-                .username("Florence")
-                .authToken(System.getenv("slack_api_token"))
-                .create());
-
-        startUpAlerter = new StartUpAlerterImpl(slackClient, slackChannelsToNotfiyOnStartUp());
+        this.startUpAlerter = initStartUpAlerter();
 
         info().data("root_path", rootPath.toString())
                 .data("zebedee_path", zebedeePath.toString())
@@ -273,6 +269,21 @@ public class ZebedeeConfiguration {
         }
 
         return centralKeyring;
+    }
+
+    private StartUpAlerter initStartUpAlerter() {
+        List<String> startUpNotificationRecipients = slackChannelsToNotfiyOnStartUp();
+
+        if (startUpNotificationRecipients == null || startUpNotificationRecipients.isEmpty()) {
+            return new NoOpStartUpAlerter();
+        }
+
+        SlackClient slackClient = new SlackClientImpl(new Profile.Builder()
+                .emoji(":flo:")
+                .username("Florence")
+                .authToken(System.getenv("slack_api_token"))
+                .create());
+        return new StartUpAlerterImpl(slackClient, slackChannelsToNotfiyOnStartUp());
     }
 
     public boolean isUseVerificationAgent() {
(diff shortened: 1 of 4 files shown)


AngelaLW (dismissed):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -275,6 +271,21 @@ private Keyring initCentralKeyring() throws KeyringException {
...
+        List<String> startUpNotificationRecipients = slackChannelsToNotfiyOnStartUp();
+
+        if (startUpNotificationRecipients == null || startUpNotificationRecipients.isEmpty()) {
+            return new NoOpStartUpAlerter();
+        }
+
+        SlackClient slackClient = new SlackClientImpl(new Profile.Builder()
+                .emoji(":flo:")
+                .username("Florence")
+                .authToken(System.getenv("slack_api_token"))
```
jessjenkins, line 284: Do we need to protect against this being null too? I guess this was the main problem as I ran the code locally successfully with a null channels list, but I already had a slack token, suggesting that it was the missing slack token rather than channel list that was Angela's issue possibly.

jessjenkins: Ah, or maybe because at that point I hadn't actually signed i to zebedee. As you were.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/NoOpStartUpAlerter.java
```diff
@@ -0,0 +1,19 @@
...
+public class NoOpStartUpAlerter implements StartUpAlerter {
+
+    @Override
+    public void queueLocked() {
+        info().log("NoOpStartUpAlerter queueLocked invoked - do nothing");
+    }
+
+    @Override
+    public void queueUnlocked() {
+        info().log("NoOpStartUpAlerter queueUnlocked invoked - - do nothing");
```
jessjenkins, line 17: Hyper-hyphenation
```suggestion
        info().log("NoOpStartUpAlerter queueUnlocked invoked - do nothing");
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/StartUpAlerterImpl.java
```diff
@@ -40,18 +50,46 @@ public StartUpAlerterImpl(SlackClient slack, List<String> channels) {
...
     public void queueLocked() {
         synchronized (MUTEX) {
             if (queueLocked && !lockedNotificationSent) {
-                channels.stream().forEach(c -> messages.add(publishingQueueLocked(c)));
+
+                List<Future<PostMessage>> futures = null;
+                try {
+                    futures = executorService.invokeAll(queueLockedAlerts);
+                } catch (Exception ex) {
+                    error().exception(ex).log("error sending publishing queue locked slack notification");
```
jessjenkins, line 77: Being pedantic, but the error is in creating the futures that will send notifications, not with sending the notifications themselves.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/StartUpAlerterImpl.java
```diff
@@ -40,18 +50,46 @@ public StartUpAlerterImpl(SlackClient slack, List<String> channels) {
...
+                    return;
+                }
+
                 this.lockedNotificationSent = true;
```
jessjenkins, line 81: Just wondering if this is needed anymore? The notification(s) are sent if `queueUnlockedAlerts` is no-empty. In any case it isn't true if all fail as this wont be identified until the futures are consumed in the next block.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/util/slack/StartUpAlerterImplTest.java
```diff
@@ -149,13 +147,10 @@ public void testNotifyLocked_noChannelsConfigured_shouldDoNothing() throws Excep
     }
 
     @Test
-    public void testNotifylocked_originalMessageNull_shouldThrowEx() throws Exception {
-        alerter = new StartUpAlerterImpl(slackClient, null, true, true);
-
-        RuntimeException ex = assertThrows(RuntimeException.class, () -> alerter.queueUnlocked());
+    public void testNotifylocked_originalMessageNull_shouldDoNothing() throws Exception {
```
jessjenkins, line 150: Rename test method maybe? We are testing notifying of unlocked
```suggestion
    public void testNotifyUnlocked_originalMessageNull_shouldDoNothing() throws Exception {
```

jessjenkins: In fact, is this test method not identical to the one at line 178?
`testNotifyUnlocked_originalMessagesNull_shouldNotSendUpdate`

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/StartUpAlerterImpl.java
daiLlew: Good point. This was a hangover from the first iteration, i'll remove it. The same can be determined by queueUnlockedAlerts being non empty.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/util/slack/StartUpAlerterImplTest.java
daiLlew: Possibly. Ill revisit the tests they're probably out of sync with the current version.

jessjenkins (approved):

PR #461, opened 2021-06-03
Added new API to list collection IDs of keys stored in users keyring

daiLlew:
### What

Created a new API endpoint to list the collection ID's for the keys a user has stored on their keyring.

**Returns the collection ID's only** not key values.


morses-code (approved):
LGTM 👍🏻

PR #466, opened 2021-06-29
stashing changes

daiLlew:
### What

Updated Teams API endpoint to use new keyring abstraction + some other minor updates.

I have written integration tests to prove the functionality is still correct. You can find these [here](https://github.com/ONSdigital/dp-zebedee-keyring-migration-tests) they cover adding/removing teams and adding removing team members.

### How to review

Check it makes sense.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -480,46 +484,101 @@ public static Collection update(
...
-                for (Team team : teamsService.listTeams()) { // iterate the teamsService list to find the team object
-                    if (currentTeamId.equals(team.getId())) { // if the ID's match
-                        if (!collectionDescription.getTeams().contains(team.getName())) { // if the team is not
-                            // listed in the updated list
-                            zebedee.getPermissionsService().removeViewerTeam(collectionDescription, team, session);
-                        }
-                    }
-                }
+
+            if (desc != null && desc.getTeams() != null) {
```
mwnicks, line 505: Is this `if` not the same as the outer one?
Maybe I'm missing something

mwnicks (approved):

PR #467, opened 2021-06-30
removed unused splunk metrics reporting service, client and models

daiLlew:
### What

Removed the redundant `Splunk` reporting service, models, client and unit tests.

This metrics reporting functionality was never used and the dependency was causing issues with the CI build so I have ripped it out.

I've left the `NoOp` implementation of the `MetricsService` for now. This service is referenced in a few places but I've updated the config to so its hardwired to always return the `NoOp` version which effectively does nothing.  If we ever need to re-implement it we would only need to create a new metrics service as its plumbed into the places that would need it.

### How to review

Make sure I have deleted the right things and the unit tests still pass.

### Who can review

@CarlHembrough @jessjenkins I would appreciate it if you could give this a sanity check.


CarlHembrough (commented):
The changes look good to me. There are some redundant splunk related env vars being set in `run.sh` and `zebedee-reader/scripts/codedeploy/application-start.sh` that could be removed.

(not sure if the code deploy stuff is being used? 🤔 )

CarlHembrough (approved):

PR #473, opened 2021-07-05
removed exception for key missing

daiLlew:
### What

Updated keyring to log error if key is missing from src user keyring.

Previously an exception was thrown but this prevents the publishing team from being able to reset user passwords. There is a separate issue here as to why a key is missing but this will be investigated separately. This change should restore the ability to reset a user password.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
 import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
 
 /**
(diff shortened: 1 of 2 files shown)


sandypadmanabhan (approved):
LGTM

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/LegacyKeyringImpl.java
```diff
@@ -349,7 +350,18 @@ public void assignTo(User src, User target, List<CollectionDescription> assignme
...
         for (CollectionDescription desc : assignments) {
             if (!srcCache.keySet().contains(desc.getId())) {
-                throw new KeyringException(CACHED_KEY_MISSING_ERR, desc.getId());
+                // If a key is not in the src users keyring cache then there is a potential issue with how the keys
+                // are assigned. We'll log this error instead of throwing it so user functionality isn't affected but
+                // this should be investigated if it occurs unexpectedly.
+                error().collectionID(desc)
+                        .data("src_user", src.getEmail())
+                        .data("target_user", target.getEmail())
+                        .log("error assiging collection key to target user. The src user is missing this collection key" +
```
HeJones, line 359: Please can I ask what is the difference between creating the error in this way to throwing a new exception.

daiLlew: `error()` creates new log event with level `Error`, it doesn't actually throw an exception.

Originally this was throwing an exception which was causing other stuff to break, so to reduce the wider impact on the rest of the service I've updated this to log the error instead of throwing it.

HeJones: Thank you, this makes sense

PR #476, opened 2021-07-05
Feature/implement key pair

HeJones:
### What

- Replace sessions_api feature flag with jwt_sessions
- Remove sessionsServiceAPI and implementations 
- Instantiate new implementation of sessions interface (JWTStore) on start up
- Constructor to create a map pair from the key id and signing key from config

### Who can review

Anyone but me


daiLlew (approved):
LGTM

mike-kelly-methods:
Me too 👍

PR #477, opened 2021-07-06
bugfix - Allows deleting visualisations

cadmiumcat:
### What

Currently there is a bug that does not allow you to delete visualisations. It tries to delete the directory for the visualisation, but the path tha is passed in is for a file (/data.json). 

This PR ensures that the visualisation folder is removed.

### How to review
Check that the code change makes sense

- run the publishing stack
- create collection --> create/edit content --> Old workspace --> visualisations (in Browse) --> click upload visualisation --> enter a unique id and page name, then create page --> Click Visualistion (file upload - it must be a zip file) then save and submit for review.
- click on the visualisation that is now awaiting for review --> Delete --> OK
If no errors appear, it's worked. 
You can double check that the folder has been deleted from `dp-zebedee-content`

### Who can review

anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -1175,7 +1175,7 @@ public class Collection {
 
         String visualisationZipUri = contentPath.getParent().toString();
         if (visualisationZipUri == null || StringUtils.isEmpty(visualisationZipUri)){
-            info().data("zip", visualisationZipUri).log("unable to delete visualisation")
+            info().data("zip", visualisationZipUri).log("unable to delete visualisation");
             return false;
         }
         String dataJsonUri = resolveDataVizDataJsonURI(contentPath);
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -1169,13 +1169,13 @@ public class Collection {
     }
 
     public boolean deleteDataVisContent(Session session, Path contentPath) throws IOException {
-        if (contentPath == null || StringUtils.isEmpty(contentPath.toString())) {
+        if (StringUtils.isEmpty(contentPath.toString())) {
             return false;
         }
 
         String visualisationZipUri = contentPath.getParent().toString();
-        if (visualisationZipUri == null || StringUtils.isEmpty(visualisationZipUri)){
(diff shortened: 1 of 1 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1173,7 +1173,7 @@ public boolean deleteDataVisContent(Session session, Path contentPath) throws IO
             return false;
         }
 
-        String visualisationZipUri = contentPath.toString();
+        String visualisationZipUri = contentPath.getParent().toString();
```
daiLlew, line 1176: This should be fine but it might be wise to check that `getParent()` does not return null or empty. 

We should also check that `visualisationZipUri` is not equal to the `collectionDir` (line 1185).

For example we don't want to delete this dir if its equal to:
- `mycollection/inprogress`
- `mycollection/complete` 
- `mycollection/reviewed`

As this would actually end up deleting the collection directory not just the content - this would be bad. If any of these are true skip the deleted.

Happy to talk through this if you want more of an explanation.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1173,7 +1173,11 @@ public boolean deleteDataVisContent(Session session, Path contentPath) throws IO
...
+        if (visualisationZipUri == null || StringUtils.isEmpty(visualisationZipUri)){
```
daiLlew, line 1177: [`StringUtils.isEmpty` ](http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-) will handle null cases so you can remove the `visualisationZipUri == null` check.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1173,7 +1173,11 @@ public boolean deleteDataVisContent(Session session, Path contentPath) throws IO
...
+            info().data("zip", visualisationZipUri).log("unable to delete visualisation");
```
daiLlew, line 1178: Maybe add a bit more context to this log message - `"unable to delete visualisation path provided was null/empty"`.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
cadmiumcat: you're right! I was following the pattern I saw in other checks, but it does seem redundant

daiLlew (approved):

PR #479, opened 2021-07-09
Add Kafka producer for contend-published events

jessjenkins:
### What

Add a kafka producer to zebedee cms. When this feature is enabled a kafka message will be produced for each uri in a collection on publish.

### How to review

Review code, ensure that nothing is produced when the feature is disabled.

### Who can review

Anyone but me


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -52,39 +47,18 @@
...
-import static com.github.onsdigital.zebedee.Zebedee.PERMISSIONS;
-import static com.github.onsdigital.zebedee.Zebedee.PUBLISHED;
-import static com.github.onsdigital.zebedee.Zebedee.PUBLISHED_COLLECTIONS;
-import static com.github.onsdigital.zebedee.Zebedee.SERVICES;
-import static com.github.onsdigital.zebedee.Zebedee.SESSIONS;
-import static com.github.onsdigital.zebedee.Zebedee.TEAMS;
-import static com.github.onsdigital.zebedee.Zebedee.USERS;
-import static com.github.onsdigital.zebedee.Zebedee.ZEBEDEE;
+import static com.github.onsdigital.logging.v2.event.SimpleEvent.*;
+import static com.github.onsdigital.zebedee.Zebedee.*;
```
daiLlew, line 59: Can you disable the Wildcard imports please. https://intellij-support.jetbrains.com/hc/en-us/community/posts/206203659-Turn-off-Wildcard-imports-

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
```diff
@@ -0,0 +1,12 @@
+package com.github.onsdigital.zebedee.kafka;
+
+import org.apache.kafka.clients.producer.RecordMetadata;
+
+import java.util.concurrent.Future;
+
+/**
+ * Interface representing a client that sends messages to kafka
+ */
+public interface KafkaClient {
```
daiLlew, line 10: Is this likely to ever have any other methods? If no can you add the 
```
@FunctionalInterface
```
Annotation please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
```diff
@@ -0,0 +1,12 @@
...
+    public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID);
```
daiLlew, line 11: Javadoc.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -149,6 +148,17 @@ public static boolean executePublish(Collection collection, CollectionReader col
...
 
+        if (CMSFeatureFlags.cmsFeatureFlags().isKafkaEnabled()) {
+            try {
+                sendToKafka(collection);
+            } catch (Exception e) {
+                error().data("collectionId", collection.getDescription().getId()).data("publishing", true)
+                        .logException(e, "failed to send content-published kafka events");
+                PostMessageField msg = new PostMessageField("Error", e.getMessage(), false);
+                collectionAlarm(collection, "Failed to send content-published kafka events", msg);
+            }
```
daiLlew, line 159: May as well move all of this inside the `sendToKafka` method.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/service/KafkaServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.github.onsdigital.zebedee.kafka.KafkaClient;
+import org.apache.kafka.clients.producer.RecordMetadata;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.*;
+import java.util.stream.Collectors;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.*;
```
daiLlew, line 11: Wildcard import.

daiLlew (commented):
Few minor tidy up comments but otherwise LGTM 👍

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
jessjenkins: I've updated the class count from 5 to 50.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
jessjenkins: It might have more methods in future if there are other producers needed. At the moment zebedee will only raise a `content-published` event but if in future it needed others (say a `user-logged-in` or something like that) then there would be more methods. I have perhaps muddied the water a little by using a functional lambda to define a no-op client when the feature is turned off, but my thought is that by that stage it would no longer need a feature flag. I could replace this with a no op class instead if it made it clearer?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -747,6 +754,18 @@ private static boolean publishDatasets(Collection collection) throws IOException
...
 
+    private static void sendToKafka(Collection collection) throws IOException {
+        List<String> uris = collection.getReviewed().uris();
+        try {
+            kafkaServiceSupplier.getService().produceContentPublished(collection.getId(), uris);
+        } catch (Exception e) {
+            error().data("collectionId", collection.getDescription().getId()).data("publishing", true)
+                    .logException(e, "failed to send content-published kafka events");
+            PostMessageField msg = new PostMessageField("Error", e.getMessage(), false);
+            collectionAlarm(collection, "Failed to send content-published kafka events", msg);
```
daiLlew, line 765: Can you pull the latest develop if you haven't already - Sandy has updated how we send slack notifications and I think this class (SlackNotification) is now deprecated. We've updated this class already so it should have everything required to use the new approach. There should be examples already but as a rough guide it looks like:
```
 String channel = Configuration.getDefaultSlackAlarmChannel();
Notifier notifier = zebedee.getSlackNotifier();
notifier.sendCollectionAlarm(collection, channel, "you message here, ex);
```

daiLlew (approved):

PR #482, opened 2021-07-12
Add check on isJwtSessionsEnabled feature flag in AuthenticationFilter.Filter() routine.

mike-kelly-methods:
### What
As per title. The sessions source changes depending on feature flag isJwtSessionsEnabled setting.

This represents both the old and new sessions model for integration of dp-identity-api to zebedee.

### How to review
Run unit tests and visually sanitise code.

### Who can review
Anyone but me.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
@@ -37,9 +37,7 @@ public class AuthenticationFilter implements PreFilter {
 
     private Sessions sessions;
 
-    private final String NO_AUTH_HEADER_FOUND       = "No authorisation header found. Exiting...";
-    private final String ACCESS_TOKEN_EXPIRED_ERROR = "JWT verification failed as token is expired.";
-    private final String TOKEN_NOT_VALID_ERROR      = "Token format not valid.";
+    private final String NO_AUTH_HEADER_FOUND = "No authorisation header found. Exiting...";
 
     private boolean jwtSessionsEnabled = false;
 
@@ -94,7 +92,6 @@ public class AuthenticationFilter implements PreFilter {
      */
     @Override
     public boolean filter(HttpServletRequest request, HttpServletResponse response) {
-
         // Pass through OPTIONS request without authentication for cross-origin preflight requests:
         if (StringUtils.equalsIgnoreCase("OPTIONS", request.getMethod())) {
             return true;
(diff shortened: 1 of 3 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
```diff
@@ -66,35 +100,69 @@ public boolean filter(HttpServletRequest request, HttpServletResponse response)
...
+                        if (e.getMessage() == ACCESS_TOKEN_EXPIRED_ERROR || e.getMessage() == TOKEN_NOT_VALID_ERROR ) {
+                            unauthorisedRequest(response, e.getMessage());
+                        } else {
+                            accessTokenError(response, e.getMessage());
+                        }
+                    }
+                }
+            } catch (IOException e) {
+                error(response);
+            }
```
morses-code, line 131: Can this be its own method, would help with the flow and readability in my opinion.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
```diff
@@ -66,35 +100,69 @@ public boolean filter(HttpServletRequest request, HttpServletResponse response)
...
+            try {
+                Session session = sessions.get(request);
+    
+                if (session == null) {
+                    forbidden(response);
+                } else {
+                    result = true;
+                }
+            } catch (IOException e) {
+                error(response);
```
morses-code, line 142: Same here possibly

morses-code (commented):
Just a few comments, but otherwise looks ok and tests are passing 👍🏻

morses-code (dismissed):
LGTM 👍🏻

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
```diff
@@ -66,18 +100,65 @@ public boolean filter(HttpServletRequest request, HttpServletResponse response)
...
+            //ensure that authorisation header present
+            if (request.getHeader("Authorization") == null) {
+                unauthorisedRequest(response, NO_AUTH_HEADER_FOUND);
+            } else {
+                try {
+                    sessions.set(request.getHeader("Authorization"));   
+                    result = true;
+                } catch (SessionsStoreException e) {
+                    // treat access token expired or malformed access token as unauthorised
+                    if (e.getMessage() == ACCESS_TOKEN_EXPIRED_ERROR || e.getMessage() == TOKEN_NOT_VALID_ERROR ) {
```
daiLlew, line 138: I think there are specific exception classes for both of these cases which is safer than doing an error message check.

```
catch (SessionsTokenExpiredException | SessionsDecodeException ex  ) {
    unauthorisedRequest(response, e.getMessage());
} catch (Exception ex) {
    // catch all logic.
   ...
}

File: zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java [hunk 4]
```diff
@@ -66,18 +100,65 @@ public boolean filter(HttpServletRequest request, HttpServletResponse response)
...
+                        unauthorisedRequest(response, e.getMessage());
+                    } else {
+                        accessTokenError(response, e.getMessage());
+                    }
+                }
+            }
+        } catch (IOException e) {
```
daiLlew, line 145: See comment below for `unauthorisedRequest`. If you implement the suggestion or something similar it will remove the need to have these try-catch blocks for this IOException.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
```diff
@@ -89,12 +170,24 @@ public boolean filter(HttpServletRequest request, HttpServletResponse response)
         return result;
     }
 
+    private void unauthorisedRequest(HttpServletResponse response, String errorMessage) throws IOException {
+        response.setContentType("application/json");
+        response.setStatus(HttpStatus.UNAUTHORIZED_401);
+        Serialiser.serialise(response, errorMessage);
+    }
```
daiLlew, line 177: If you do something like this (unAuth response by default to error response if unsuccessful):
```
    private void unauthorisedRequest(HttpServletResponse response, String errorMessage) {
        try {
            response.setContentType("application/json");
            response.setStatus(HttpStatus.UNAUTHORIZED_401);
            Serialiser.serialise(response, errorMessage);
        } catch (IOException ex) {
            error(response);
        }
    }
```
Then you won't have to catch the thrown IOException multiple times in the `verifyAndStoreAccessToken`. You can then remove the try catch blocks and it will more straight forward.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
mike-kelly-methods: There are - I'll use those.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/filters/AuthenticationFilter.java
mike-kelly-methods: 👍

daiLlew (approved):

PR #487, opened 2021-07-14
Fix potential infinite loop when logging collection

jessjenkins:
### What

Fixed issue with a collection object being passed into the logger and causing an infinite loop when serialising to JSON.

### How to review

Ensure code makes sense and tests pass.

### Who can review

Anyone but me


HeJones (commented):
This makes sense to me but I think it would be good if someone else looked at it

daiLlew (approved):

PR #491, opened 2021-07-15
stipping out unused csdb import code

daiLlew:
### What
More keyring migration work.

The csdb import functionality was added but never used. This code is considerably out of date and relies on the legacy keyring. Rather than refactor this to use the new central keyring I am deleting it - simpler.


mike-kelly-methods (approved):
👍

PR #494, opened 2021-07-16
removed application keys functionality - this is old and unused

daiLlew:
### What

Remove the application keys functionality. This is old, unused and in the way - bin it! 

I've deleted `ApplicationKeys` from the code and deleted all uses of it. As stated this was never used so I do not anticipate any issues from it being removed.


CarlHembrough (changes requested):
Changes look fine - though I think there is some additional code that can be removed:

- both classes in the com.github.onsdigital.zebedee.json.encryption namespace: `StoredApplicationKeys` and `StoredKeyPair`
- a `initialiseCsdbImportKeys` function commented out in `Root.java` on line 148

CarlHembrough (approved):

PR #497, opened 2021-07-23
fix delete zip defect

daiLlew:
### What

Fix defect deleting data viz pages/zips.

Deleting standard content from the CMS requires a path to the file being deleted - a `data.json` file . However, data viz content is a special case and requires a different approach.

When deleting a data viz page or zip file we need to delete a directory not just a single file so the URI need to be for the **parent directory of the content being deleted**. 

This fix updates the logic generating the filepath to delete if the path points to a data.json file we return the parent dir instead (i.e. the directory the data.json lives in).

Example: 
If URI is `/a/b/c/data.json` the path to delete is `/a/b/c`.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/DataVisualisationZip.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/DataVisualisationZip.java
@@ -150,7 +150,7 @@ public class DataVisualisationZip {
             throw new BadRequestException(NO_ZIP_PATH_ERROR_MSG);
         }
 
-        info().data("zipPath", zipPath).log(DELETING_ZIP_DEBUG);
+        info().data("zipPath", zipPath).log(UNZIP_DEBUG);
 
         Session session = zebedeeCmsService.getSession(request);
         com.github.onsdigital.zebedee.model.Collection collection = zebedeeCmsService.getCollection(request);
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/DataVisualisationZip.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/DataVisualisationZip.java
@@ -29,6 +29,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/DataVisualisationZip.java
```diff
@@ -180,16 +181,19 @@ private void unzipContent(ContentWriter contentWriter, Resource zipRes, String z
...
             Path filePath;
 
+            List<String> filesWritten = new ArrayList<>();
             while (zipEntry != null) {
 
                 if (isValidDataVisContentFile.test(zipEntry)) {
                     filePath = zipDir.resolve(zipEntry.getName());
                     contentWriter.write(zipInputStream, filePath.toString());
-                    info().data("zipPath", zipPath).log(DELETING_ZIP_DEBUG);
+                    filesWritten.add(filePath.toString());
```
jessjenkins, line 190: I see you've removed the misleading `DELETING_ZIP_DEBUG` log line, but there's another one at L153 that I think could be changed too.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1167,16 +1167,20 @@ public boolean deleteDataVisContent(Session session, Path contentPath) throws IO
             return false;
         }
 
-        String visualisationZipUri = contentPath.toString();
+        String visualisationZipUri = resolveDataVizZipURI(contentPath);
```
jessjenkins, line 1170: Question on naming. The function mat be for the purpose of data viz zips but in reality it's just the parent dir if the file to delete is `data.json`. Maybe something like `parentDir` or something?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -1203,6 +1207,30 @@ private String resolveDataVizDataJsonURI(Path contentPath) {
...
+     * <p>
+     * This method will return a path to the parent dir if the uri is a data.json file.
+     * Example:
+     * If uri is "a/b/c/data.json" then it will return "a/b/c".
+     *
+     * If uri is "a/b/c" then it will return "a/b/c".
+     *
+     * @param uri the file path to check.
+     */
+    private String resolveDataVizZipURI(Path uri) {
```
jessjenkins, line 1225: Following on from other naming comment, I just wonder if this method is really about a zip dir? In the context it is being used it is, but in isolation this is a method that finds a data json's parent dir unless that dir is root.

jessjenkins (approved):

PR #503, opened 2021-08-09
Feature/delete keys

daiLlew:
### What

Updated `DELETE /collection` endpoint to also remove key from keyring.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
@@ -4,12 +4,14 @@ import com.github.davidcarboni.restolino.framework.Api;
 import com.github.onsdigital.zebedee.audit.Audit;
 import com.github.onsdigital.zebedee.exceptions.BadRequestException;
 import com.github.onsdigital.zebedee.exceptions.ConflictException;
+import com.github.onsdigital.zebedee.exceptions.InternalServerError;
 import com.github.onsdigital.zebedee.exceptions.NotFoundException;
 import com.github.onsdigital.zebedee.exceptions.UnauthorizedException;
 import com.github.onsdigital.zebedee.exceptions.ZebedeeException;
 import com.github.onsdigital.zebedee.json.CollectionDescription;
 import com.github.onsdigital.zebedee.json.CollectionType;
 import com.github.onsdigital.zebedee.keyring.Keyring;
+import com.github.onsdigital.zebedee.keyring.KeyringException;
 import com.github.onsdigital.zebedee.keyring.KeyringUtil;
 import com.github.onsdigital.zebedee.permissions.service.PermissionsService;
 import com.github.onsdigital.zebedee.session.model.Session;
@@ -30,6 +32,7 @@ import java.io.IOException;
 import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
 import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
 import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
+import static java.text.MessageFormat.format;
 import static org.apache.commons.lang3.StringUtils.trim;
 
 
@@ -283,12 +286,7 @@ public class Collection {
             throw new NotFoundException("The collection you are trying to delete was not found");
         }
 
-        collections.delete(collection, session);
-
-        scheduleCanceller.cancel(collection);
-
-        User user = KeyringUtil.getUser(usersService, session.getEmail());
-        keyring.remove(user, collection);
+        deleteCollection(collection, session);
 
         Audit.Event.COLLECTION_DELETED
                 .parameters()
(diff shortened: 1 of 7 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -317,4 +356,17 @@ private void requireEditPermission(String email) throws UnauthorizedException {
             throw new UnauthorizedException("You are not authorised to edit collections.");
         }
     }
+
+    /**
+     * ScheduleCanceller is an abstraction wrapper for functionality to cancel a scheduled publish.
+     */
+    public static interface ScheduleCanceller {
```
sandypadmanabhan, line 363: Not sure why the interface has been added to the same file. We don't normally do this?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/CollectionTest.java
```diff
@@ -246,4 +269,197 @@ public void testPost_permissionDenied_shouldThrowException() throws Exception {
...
+        when(usersService.getUserByEmail(TEST_EMAIL))
+                .thenReturn(user);
+
+        boolean result = endpoint.deleteCollection(mockRequest, mockResponse);
+
+        assertTrue(result);
+        verify(collections, times(1)).delete(collection, session);
+        verify(scheduleCanceller, times(1)).cancel(collection);
+        verify(keyring, times(1)).remove(user, collection);
+    }
```
sandypadmanabhan, line 464: Do we need to add test for schedule canceller too?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -285,6 +301,29 @@ public boolean deleteCollection(HttpServletRequest request, HttpServletResponse
         return true;
     }
 
+    private void deleteCollection(com.github.onsdigital.zebedee.model.Collection c, Session s)
+            throws InternalServerError, NotFoundException, BadRequestException {
```
sandypadmanabhan, line 305: What about unauthorised error?

sandypadmanabhan: Does it fall under BadRequestException?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -285,6 +301,29 @@ public boolean deleteCollection(HttpServletRequest request, HttpServletResponse
...
+        // Delete the collection.
+        try {
+            collections.delete(c, s);
+        } catch (Exception ex) {
+            String message = format("error attempting to delete collection: {0}", c.getId());
+            throw new InternalServerError(message, ex);
+        }
+
+        // Cancel any scheduled publish for this collection.
+        scheduleCanceller.cancel(c);
```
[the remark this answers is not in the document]
daiLlew: No. The actual work for cancelling a schedule publish is done inside of `Root.cancelPublish()` (this is an existing method). That method swallows and logs any exception should it occur.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
daiLlew: This is the only place this interface is/will ever but used so I created it within this class instead of in its own file.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/CollectionTest.java
daiLlew: The actual cancel functionality is done in Root and is existing functionality. I've added an interface abstraction around it to make it easier to test this class. I've taken the decision to not test that code as its beyond the scope of the change I am making here.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
daiLlew: None of the methods called here will thrown an Unauthorised exception so it doesn't need to be added to the method signature.

sandypadmanabhan: But when the session is null, it throws an unauthorised exception according to the tests....

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -285,6 +301,36 @@ public boolean deleteCollection(HttpServletRequest request, HttpServletResponse
...
+
+        // Remove the collection encryption key from the keyring
+        User user = null;
+        try {
+            user = KeyringUtil.getUser(usersService, s.getEmail());
+        } catch (Exception ex) {
+            String message = format("error attempting to get user from session details: {0}", s.getEmail());
+            throw new InternalServerError(message, ex);
```
sandypadmanabhan, line 323: when we use the try catch block for get user all the exception get's caught as internal server error. In this case do we still need NotFoundException and BadRequestException defined in the function signature?

sandypadmanabhan (approved):
when we use the try catch block for get user all the exception get's caught as internal server error. In this case do we still need NotFoundException and BadRequestException defined in the function signature?

PR #506, opened 2021-08-11
New dp-logging library update

bpathak-ons:
### What

new dp-logging released version update

### How to review

new dp-logging released version - v2.0.0-beta

### Who can review

anyone not me


DavidSubiros (commented):
The changes look good to me, but you may need to check some CI steps that filed (maybe you just need to re-trigger them)

rafahop (approved):
Looks good to me.
I see `dp-logging` uses jackson version 2.12.1 and it is a transitive dependency hence the need of bumping it here so they match and we avoid conflicts 👍

PR #508, opened 2021-08-17
stashing changes

daiLlew:
### What

Added new `MigrationSchedulerKeyCache` `SchedulerKeyCache` implementation. Like the other keyring migration classes this dual runs the new and legacy implementations allowing us to migrate to the new world smoothly.

Unit tests all passing. I've also testing this locally with manual an scheduled publishes:
- New keyring feature flag **off** 
    - Collection access and publishing works as expected. 
    - Admin user must login to unlock the publishing queue and publishing any pending scheduled collections.
- New keyring feature flag **on** 
    - Collection access and publishing works as expected. 
    - Restarting the CMS when a schedule publish is approved and waiting in the queue 
        - The collection is now published regardless of an admin user logging in as the new/migration keyring does not require a user to unlock the collection keyring.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/KeyringCache.java
```diff
@@ -11,7 +8,7 @@
  * {@link com.github.onsdigital.zebedee.model.Collection}s. Provides methods to add, retrieve and remove keys to/from
  * the keyring.
  */
-public interface KeyringCache {
+public interface KeyringCache extends SchedulerKeyCache {
```
CarlHembrough, line 11: The name `KeyringCache` seems a bit more generic than `SchedulerKeyCache`. The `SchedulerKeyCache` doesn't seem to have anything specific to scheduling. Consider if the naming is right for these interfaces.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/MigrationSchedulerKeyCache.java
```diff
@@ -0,0 +1,68 @@
...
+        this.centralCache = centralCache;
+        this.migrationEnabled = migrationEnabled;
+    }
+
+    @Override
+    public SecretKey get(String collectionID) throws KeyringException {
+        if (migrationEnabled) {
+            SecretKey key = getFromCentralKeyringCache(collectionID);
+            if (key == null) {
+                return key;
```
CarlHembrough, line 44: If the key is null here, shouldn't it be then attempting to read from the legacyCache? Alternatively, change the logic so that if it's not null it can return key.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/MigrationSchedulerKeyCache.java
```diff
@@ -0,0 +1,68 @@
...
+ * MigrationSchedulerKeyCache is a SchedulerKeyCache implementation that wraps around the legacy and
+ * central caches allowing us to run them both in parallel so we can migrate smoothly to the new world.
+ * <p>
+ * If migration is enabled then MigrationSchedulerKeyCache will attempt to read/write from the new central scheduler
+ * cache falling back on the legacy implementation if that is unsuccessful. If migration is not enabled - it
+ * reads/writes to/from the legacy scheduler caches only
+ * <p>
+ * This is a temp class that will be deleted once the migration to new world keyring is complete.
+ */
+public class MigrationSchedulerKeyCache implements SchedulerKeyCache {
```
CarlHembrough, line 19: Consider adding unit tests for this class

CarlHembrough (changes requested):

PR #511, opened 2021-08-19
renamed Keyring interface and updated uses of it

daiLlew:
### What

I've renamed the new `Keyring` interface to `CollectionKeyring` to prevent confusion with the existing legacy `Keyring` class - this is a large and complex change and the clashing names was making very difficult to understand.

There should be no functional change at all - its just a rename.


morses-code (approved):
LGTM

PR #514, opened 2021-08-23
Fix schedule cache for central keyring

daiLlew:
### What

Fixed schedule cache config for new central keyring. Previously scheduled publishes were unable to retrieve collection keys from the scheduler cache when the central keyring feature was enabled.


morses-code (approved):
LGTM

PR #516, opened 2021-08-23
updated Zebedee to use No op alternatives for Slack integration when …

daiLlew:
### What
Added no op alternatives for the Slack integration classes. These will be used when the Slack config is not available on start up.

No op `SlackClient`  and `Notifier` will log the messages but no post requests to the Slack API.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/NopSlackClientImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/NopSlackClientImpl.java
@@ -1,0 +1,31 @@
+package com.github.onsdigital.zebedee.util.slack;
+
+import com.github.onsdigital.slack.Profile;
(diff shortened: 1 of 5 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/slack/NopSlackClientImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * No op implementation of {@link SlackClient}
+ */
+public class NopSlackClientImpl implements SlackClient {
+
+    @Override
+    public PostMessageResponse sendMessage(PostMessage postMessage) {
+        info().data("message", postMessage.getText()).log("NopSlackClientImpl.sendMessage");
+        return null;
```
CarlHembrough, line 18: I think returning null in these methods could cause a null pointer exception, as the return value is used in some cases. Consider if this will cause any problems.

CarlHembrough (approved):

PR #517, opened 2021-08-24
Feature/remove key on publish

daiLlew:
### What

Updated POST-PUBLISH task to remove the collection key from the collection keyring. Also updated logging calls to use the `collectionID()` method instead of `data(k, v).

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PostPublisher.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PostPublisher.java
@@ -47,8 +47,9 @@ import java.util.TimeZone;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.info;
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.error;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
 import static com.github.onsdigital.zebedee.persistence.CollectionEventType.COLLECTION_POST_PUBLISHED_CONFIRMATION;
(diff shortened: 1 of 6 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PostPublisher.java
```diff
@@ -47,8 +47,9 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.info;
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.error;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
```
bppathak, line 51: changing the logging lib, is this correct ?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PostPublisher.java
```diff
@@ -105,14 +106,13 @@ public static boolean postPublish(Zebedee zebedee, Collection collection, boolea
...
+            zebedee.getSchedulerKeyCache().remove(collection.getId());
 
-            // FIXME using PostPublisher.getPublishedCollection feels a bit hacky
             SlackNotification.publishNotification(publishedCollection,SlackNotification.CollectionStage.POST_PUBLISH, SlackNotification.StageStatus.COMPLETED);
 
             return true;
         } catch (Exception exception) {
-            error().data("collectionId", collection.getDescription().getId()).logException(exception, "An error occurred during the publish cleanup");
-            // FIXME using PostPublisher.getPublishedCollection feels a bit hacky
+            error().collectionID(collection).exception(exception).log("An error occurred during the publish cleanup");
```
bppathak, line 115: To be specific in logs, should we not capture collectionID for every exception handling even for null

bpathak-ons: ignoring this comment in assumption of traceID in logs are out of scope of this PR.

bpathak-ons (approved):

PR #521, opened 2021-09-06
Feature/update migration get

daiLlew:
### What

Updated `MigrationCollectionKeyring`:
- If `migrationEnabled == true` it will attempt to get the key from the collectionKeyring falling back on the legacy keyring if the requested key is not found.
- If `migrationEnabled == false` it will attempt to retrieve the key from the legacy keyring only.
- Updated the logic to catch and handle `KeyNotFoundException`s - this should return null instead of erroring.
- Split the test into a separate class per method as the tests were getting pretty large - hopefully this makes it easier to read.


morses-code (approved):
LGTM

PR #523, opened 2021-09-20
Updated listKeyring endpoint to accept src param

daiLlew:
### What

- Updated `ListKeyring` endpoint to accept src param to allow the caller to specify which keyring to use.
- Updated CMS config to use the migration keyring by default. The CMS will now write/remove from both keyring impls in parallel.

If the central keyring feature flag is enabled then get key requests will attempt to read from the central keyring first and default to the legacy impl if the key is not found.
If the feature flag is disabled then write/remove keys will update both BUT get requests will only use the legacy keyring.

### How to review

Code review. I have separate integration tests to prove this functionality is correct.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/ListKeyring.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/ListKeyring.java
@@ -1,6 +1,7 @@
 package com.github.onsdigital.zebedee.api;
 
 import com.github.davidcarboni.restolino.framework.Api;
+import com.github.onsdigital.zebedee.configuration.CMSFeatureFlags;
 import com.github.onsdigital.zebedee.exceptions.BadRequestException;
 import com.github.onsdigital.zebedee.exceptions.InternalServerError;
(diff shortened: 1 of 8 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/ListKeyring.java
```diff
@@ -63,8 +70,12 @@ public ListKeyring(final CollectionKeyring collectionKeyring, final Sessions ses
     @GET
     public Set<String> listUserKeys(HttpServletRequest request, HttpServletResponse response) throws ZebedeeException {
         checkPermission(getSession(request));
+        String src = request.getParameter("src");
+
+        CollectionKeyring keyringSrc = "central".equalsIgnoreCase(src) ? legacyKeyring : centralKeyring;
```
CarlHembrough, line 75: This looks like it may be doing the inverse of what it should. If it matches `central`, then it's using the legacyKeyring?

CarlHembrough (changes requested):
Please see comment about the new src param. Also the unit tests are failing in the pipeline.

CarlHembrough (approved):

PR #530, opened 2021-09-29
Fix for integration error

bpathak-ons:
### What

* Strip off data.json

### How to review

* look if code make sense
* build, audit and unit pass

### Who can review

* anyone not me

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
@@ -11,7 +11,7 @@ public interface KafkaClient {
 
     /**
      * Produce a 'content-published' kafka message for the given parameters
-     * @param url The url of the published resource
+     * @param uri The url of the published resource
      * @param dataType The data type of the published resource
      * @param collectionID the ID of the collection being published
      * @return
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
@@ -16,5 +16,5 @@ public interface KafkaClient {
      * @param collectionID the ID of the collection being published
      * @return
      */
-    public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID);
+    public Future<RecordMetadata> produceContentPublished(String uri, String dataType, String collectionID);
 }
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
@@ -36,7 +36,11 @@ public class KafkaClientImpl implements KafkaClient {
      */
     @Override
     public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID) {
-        String uri = url.substring(0, url.length() - "/data.json".length());
+
+        String uri = url;
+        if (url.endsWith("/data.json")) {
+            uri = url.substring(0, url.length() - "/data.json".length());
+        }
         ContentPublished value = new ContentPublished(uri,dataType,collectionID);
         return producer.send(new ProducerRecord<>(topic,value));
     }
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
@@ -36,7 +36,8 @@ public class KafkaClientImpl implements KafkaClient {
      */
     @Override
     public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID) {
-        ContentPublished value = new ContentPublished(url,dataType,collectionID);
+        String uri = url.substring(0, url.length() - "/data.json".length());
+        ContentPublished value = new ContentPublished(uri,dataType,collectionID);
         return producer.send(new ProducerRecord<>(topic,value));
     }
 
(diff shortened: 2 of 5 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
```diff
@@ -36,7 +36,8 @@ public KafkaClientImpl(String kafkaAddr, String topic) {
      */
     @Override
     public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID) {
-        ContentPublished value = new ContentPublished(url,dataType,collectionID);
+        String uri = url.substring(0, url.length() - "/data.json".length());
```
nshumoogum, line 39: Will the url always have a suffix of `/data.json`?

If not then wouldn't this remove the length of `/data.json` and could result in throwing an error/exception?

nshumoogum (changes requested):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
bpathak-ons: This will remove the suffix of /data.json, if exists, and it should be fine.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
```diff
@@ -36,7 +36,12 @@ public KafkaClientImpl(String kafkaAddr, String topic) {
      */
     @Override
     public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID) {
-        ContentPublished value = new ContentPublished(url,dataType,collectionID);
+
```
jessjenkins, line 39: I wouldn't modify the URL in this class. This class is concerned with sending an event not the content of the event. The calling class(es) would be more appropriate.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClient.java
```diff
@@ -16,5 +16,5 @@
      * @param collectionID the ID of the collection being published
      * @return
      */
-    public Future<RecordMetadata> produceContentPublished(String url, String dataType, String collectionID);
+    public Future<RecordMetadata> produceContentPublished(String uri, String dataType, String collectionID);
```
jessjenkins, line 19: Don't forget to update the JavaDoc above too (`@param url`)

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/kafka/KafkaClientImpl.java
bpathak-ons: suggested changes implemented

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -772,6 +773,18 @@ private static void sendToKafka(Collection collection) throws IOException {
         }
     }
 
+    //Removing data.json from uris, if exists
+    public static List<String> getUrlsConverted(List<String> uris) {
```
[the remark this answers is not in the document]
bpathak-ons: code updated for single uri

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -760,8 +760,13 @@ private static boolean publishDatasets(Collection collection) throws IOException
 
     private static void sendToKafka(Collection collection) throws IOException {
         List<String> uris = collection.getReviewed().uris();
+        List<String> validUris = uris.stream().map (temp -> {
+            convertUriForEvent(temp);
+            return temp;
+        }).collect(Collectors.toList());
```
jessjenkins, line 766: This isn't going to do what you think it will. The lambda is calling the conversion method but then returning the original uri rather than the converted string. The original code I suggested is simpler and should work instead.
```suggestion
        List<String> uris = collection.getReviewed().uris()
                .stream().map((temp) -> convertUriForEvent(temp))
                .collect(Collectors.toList());
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -772,6 +777,12 @@ private static void sendToKafka(Collection collection) throws IOException {
         }
     }
 
+    public static String convertUriForEvent(String uri){
```
jessjenkins, line 780: Does this method need to be public? If so add some JavaDoc to explain this method. Alternatively could it be private?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/publishing/PublisherTest.java
```diff
@@ -0,0 +1,50 @@
...
+        String testUri = "/testUri0/data.json";
+
+        //When {sending a uris to kafka}
+        String actuals = publisher.convertUriForEvent(testUri);
+
+        System.out.println(actuals);
+
+        //Then {uri does not have "data.json" string}
+        assertTrue(!actuals.contains("/testUri0/data.json"));
+        assertTrue(actuals.contains("/testUri0"));
```
jessjenkins, line 33: Use assertFalse rather than asserting `!` is true. But also is it needed anyway. You could just use assertEquals instead
```suggestion
        assertEquals("/testUri0",actuals);
```
NB. I'd rename `actuals` to `actual` too. 
PS. Same for other test

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
bpathak-ons: adding javadoc because the unit test is on this method so keeping it as public

jessjenkins: If it's only for accessibility by the unit test, you could resolve this by making it `protected` instead of `public`. I don't mind it being public if there is good reason for it to be made available to other packages but I can't see what that is. I won't object though as I see other methods in the class are also public when they may well have been better as private or protected.

jessjenkins (approved):

nshumoogum (approved):

PR #548, opened 2021-11-10
delete legacy & migration keyring code and tests

daiLlew:
### Remove the legacy/migration Keyring.

- Removed all code & test related to the legacy keyring
- Removed all code & tests related to the migration keyring.

The new keyring handles all access logic internally so part of this change is to strip out any places that were previous implementing this logic outside of the legacy keyring.

### How to review

Unit tests pass and my local integration tests pass. I need to get this into develop to continue testing.

### Who can review

Anyone.


justinpjose (approved):
Mate, this is pure gold. Less code in zebedee is beauty in my eyes. The tests all pass. I checked if the removed functions are not being used in other places and this is true

PR #549, opened 2021-11-11
fixed defect with collection keyring permission check

daiLlew:
### What

Updated `collectionKeyring.get()` permissions to allow access to any user with `view` permission.


morses-code (approved):
LGTM

PR #550, opened 2021-11-15
Release/2.0.0

daiLlew:
### Release 2.0.0

New central keyring. v2.0.0 removes all legacy collection keyring and migration keyring functional from Zebedee replacing it with the new central keyring implementation. 
- Collection keys are now stored in a central keystore instead of assigned to/removed from individual users.
- The permissions service is used to determined if a user can access a collection.
- The publishing scheduler no longer requires an admin user to login and unlock the publishing queue after a restart.

🚧 ⚠️ 
**This is a breaking change** that cannot be rolled back once it is deployed. Team 404 has discussed this and agreed that once its rolled out to prod any issues will just have to be fixed (no pressure).
⚠️ :🚧 

### How to review
🤞 
I've fixed/updated all units tests & have a separate integration test suite that I've been running against my local build and develop proving the keyring functionality works as expected. I've also manually tested several features in develop. However, this is a huge change and something could still go wrong.

Please feel free to question any changes made in this PR.

### Who can review

Anyone.


morses-code (approved):
LGTM 👍🏻

PR #551, opened 2021-11-26
Add published index endpoint for search reindexing

jessjenkins:
### What

Add a Published Index endpoint to Zebedee
- Does not need authentication as it only returns uris of published data
- Does not currently implement paging, although the model allows for paging support to be added in future.

### How to review

Check endpoint only returns published content etc.

### Who can review

Anyone but me

### DIFF
--- a/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
+++ b/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
@@ -35,7 +35,7 @@ public class PublishedIndexResponse {
                 .map(d -> new Item(d.getUri()))
                 .forEach(d -> this.items.add(d));
 
-        count += docs.size();
+        count = this.items.size();
     }
 
     public int getCount() {
--- a/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
+++ b/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
@@ -1,0 +1,87 @@
+package com.github.onsdigital.zebedee.reader.api.bean;
+
+import com.github.onsdigital.zebedee.search.indexing.Document;
+import com.google.gson.annotations.SerializedName;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Object representing a response for the PublishedIndex endpoint
+ */
+public class PublishedIndexResponse {
+
+    private int count;
+    private List<Item> items;
+    private int limit;
+    private int offset;
+    @SerializedName("total_count")
+    private int totalCount;
+
+    public PublishedIndexResponse() {
+        this.items = new ArrayList<>();
+    }
+
+    /**
+     * Adds a list of documents to the response as Items. Also updates the
+     *
+     * @param docs  List of Documents to add to response
+     */
+    public void addDocuments(List<Document> docs) {
+        if (docs == null) {
+            return;
+        }
+        docs.stream()
+                .map(d -> new Item(d.getUri()))
+                .forEach(d -> this.items.add(d));
+
+        count += docs.size();
+    }
+
+    public int getCount() {
+        return count;
+    }
+
+    public List<Item> getItems() {
+        return items;
+    }
+
+    public int getLimit() {
+        return limit;
+    }
+
+    public void setLimit(int limit) {
+        this.limit = limit;
+    }
+
+    public int getOffset() {
+        return offset;
+    }
+
+    public void setOffset(int offset) {
+        this.offset = offset;
+    }
+
+    public int getTotalCount() {
+        return totalCount;
+    }
+
+    public void setTotalCount(int totalCount) {
+        this.totalCount = totalCount;
+    }
+
+    /**
+     * Sub-object representing an individual document returned by the PublishedIndex
+     */
+    public class Item {
+        private String uri;
+
+        public Item(String uri) {
+            this.uri = uri;
+        }
+
+        public String getUri() {
+            return uri;
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


Marianasilvago (dismissed):
LGTM

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public void addDocuments(List<Document> docs) {
+        if (docs == null) {
+            return;
+        }
+        docs.stream()
+                .map(d -> new Item(d.getUri()))
+                .forEach(d -> this.items.add(d));
+
+        count += docs.size();
```
goofballLogic, line 38: shouldn't matter, but would it be slightly more robust just do `count = this.items.size();` here?

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public int getCount() {
+        return count;
+    }
+
+    public List<Item> getItems() {
+        return items;
```
goofballLogic, line 46: is it ok that we're exposing the (mutable) internal list here?

zebedee-reader/src/test/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponseTest.java
```diff
@@ -0,0 +1,84 @@
...
+public class PublishedIndexResponseTest  {
+
+    static final String URI1 = "/uri/1";
+    static final String URI2 = "/uri/2";
+
+    @Test
+    public void testAddDocument_null() {
+        PublishedIndexResponse publishedIndexResponse = new PublishedIndexResponse();
+        publishedIndexResponse.addDocuments(null);
+        assertEquals(publishedIndexResponse.getCount(),0);
```
goofballLogic, line 21: with JUnit the `expected` should come first iirc
```
    public static void assertEquals(Object expected, Object actual)
```

zebedee-reader/src/test/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponseTest.java
```diff
@@ -0,0 +1,84 @@
...
+        assertNotNull(publishedIndexResponse.getItems().get(1));
+        assertEquals(publishedIndexResponse.getItems().get(1).getUri(),URI2);
+    }
+
+    private List<Document> listOfDocuments(String... uris) {
+        return Arrays.stream(uris)
+                .map(u -> new Document(u,null))
+                .collect(Collectors.toList());
+    }
+}
```
goofballLogic, line 84: * no tests for the ability to set/get limit, offset or totalCount
* no test for serialized name of totalCount

goofballLogic: was able to get tests passing with the following implementation:
```
package com.github.onsdigital.zebedee.reader.api.bean;

import com.github.onsdigital.zebedee.search.indexing.Document;
import com.google.gson.annotations.SerializedName;

import java.util.ArrayList;
import java.util.List;

/**
 * Object representing a response for the PublishedIndex endpoint
 */
public class PublishedIndexResponse {

    private int count;
    private List<Item> items;
//    private int limit;
//    private int offset;
//    @SerializedName("total_count")
//    private int totalCount;

    public PublishedIndexResponse() {
        this.items = new ArrayList<>();
    }

    /**
     * Adds a list of documents to the response as Items. Also updates the
     *
     * @param docs  List of Documents to add to response
     */
    public void addDocuments(List<Document> docs) {
        if (docs == null) {
            return;
        }
        docs.stream()
                .map(d -> new Item(d.getUri()))
                .forEach(d -> this.items.add(d));

        count += docs.size();
    }

    public int getCount() {
        return count;
    }

    public List<Item> getItems() {
        return items;
    }
//
//    public int getLimit() {
//        return limit;
//    }
//
    public void setLimit(int limit) {
        //this.limit = limit;
    }
//
//    public int getOffset() {
//        return offset;
//    }
//
    public void setOffset(int offset) {
        //this.offset = offset;
    }
//
//    public int getTotalCount() {
//        return totalCount;
//    }
//
    public void setTotalCount(int totalCount) {
//        this.totalCount = totalCount;
    }
//
    /**
     * Sub-object representing an individual document returned by the PublishedIndex
     */
    public class Item {
        private String uri;

        public Item(String uri) {
            this.uri = uri;
        }

        public String getUri() {
            return uri;
        }
    }
}
```

zebedee-reader/src/test/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponseTest.java
jessjenkins: Thanks, you're totally right. I've been writing too much Go recently. Will update.

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
jessjenkins: Yep, that's probably a little simpler 👍

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponse.java
jessjenkins: This is a good point, but not something we tend to do currently within Zebedee. I'm goin to leaave this as is.

zebedee-reader/src/test/java/com/github/onsdigital/zebedee/reader/api/bean/PublishedIndexResponseTest.java
jessjenkins: We don't test getters and setters. But I've added a test for the serialisation.

franmoore05 (approved):
LGTM

PR #552, opened 2021-12-07
Feature/jwt permissionsservice

AWitcherONS:
### What

Added PermissionsService proxy for JWTSessionsEnabled with unit tests
added JWTPermissionsServiceImpl to override legacy methods with new methods for JWT Permissions 

### How to review

Code meets current standards is sensible and maintainable 
unit tests pass 
**_note_** there is an issue with 
Tests in error: 
  com.github.onsdigital.zebedee.keyring.central.CollectionKeyCacheImplTest: modifiers 
  comment out lines 47-49 
  mvn clean package  -Dossindex.skip=true
### Who can review
anyone but me with zebedee and dp-identity-api understanding


Marianasilvago (approved):
LGTM

Marianasilvago (dismissed):
LGTM

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
+
+public class JWTPermissionsServiceImpl implements PermissionsService {
```
daiLlew, line 20: Javadoc on class. Doesn't need to be hugely detained just outline that this implementation uses the JWT sessions implementation.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+            "error accessing JWTPermissions Service";
+    private Sessions sessionsService;
+    private Session session;
+
+
+    /**
+     * @param sessionService
+     */
+
+    public JWTPermissionsServiceImpl(Sessions sessionService) {
```
daiLlew, line 33: Complete javadoc.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * @param email the email of the user to check.
+     * @return boolean
+     * @throws IOException
+     */
+    @Override
+    public boolean isPublisher(String email) throws IOException {
+        boolean result = true;
+        if (StringUtils.isEmpty(email)) {
```
daiLlew, line 57: Logic is fine but I think we could refactor this to simplify it a bit and remove some of the nesting and else block:

```
if (StringUtils.isEmpty(email)) {
    warn().user(email).log("request for isPublisher unsuccessful empty email");
    return false;
} 

Session s = null;
try {
    Session s = getSessionfromEmail(email);
} catch (Exception ex) {
    warn().user(email).exception(ex).log("request for publisher privilages exception caught");
    returen false;
}

 return hasPermission(s, PUBLISHER_PERMISSIONS);
}
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+                }
+
+            } catch (Exception exception) {
+                warn().user(email)
+                        .log("request for admin privilages exception caught");
+                result = false;
+            }
+        }
+
+        return result;
```
daiLlew, line 117: See previous refactor example to streamline this.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+    /**
+     * @param collection the collection to check users against.
+     * @return
+     * @throws JWTVerificationException
+     */
+
+    @Override
+    public List<User> getCollectionAccessMapping(Collection collection) throws JWTVerificationException {
+        throw new JWTVerificationException(JWTPERMISSIONSSERVICE_ERROR);
+    }
```
daiLlew, line 130: Can we specifically state in the meth Javadoc that  this functionality is not supported by this implementation and it will throw an exception if called.

Can we do this for every method that is not supported please - its important to document it for future users to know it was intentional.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+
+    @Override
+    public boolean canEdit(String email) throws JWTVerificationException {
+        throw new JWTVerificationException(JWTPERMISSIONSSERVICE_ERROR);
+    }
+
+    @Override
+    public boolean canEdit(User user) throws JWTVerificationException {
+        throw new JWTVerificationException(JWTPERMISSIONSSERVICE_ERROR);
+    }
```
daiLlew, line 177: I could be wrong but I think we will need to implement all of the `canEdit` & `canView` methods as I am pretty sure these are needed internally within Zebedee.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    /**
+     * @param session {@link Session} to get the user details from.
+     * @return
+     * @throws IOException
+     */
+    @Override
+    public boolean isPublisher(Session session) throws IOException {
+        if (jwtSessionsEnabled) return jwtPermissionsService.isPublisher(session);
```
daiLlew, line 45: Can we update this to include `{`after the `if` statement please - Im not a fan of this style and we should be consistent with the rest of the codebase.

```
if (jwtSessionsEnabled) {
    return jwtPermissionsService.isPublisher(session);
}
return legacyPermissionsService.isPublisher(session);
```

daiLlew (changes requested):

daiLlew (changes requested):
I've added some comments to address - also the last commit is not verified so the build wont pass.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
+
+public class JWTPermissionsServiceImpl implements PermissionsService {
+    static final String PUBLISHER_PERMISSIONS = "publisher";
+    static final String ADMIN_PERMISSIONS = "admin";
+    static final String JWTPERMISSIONSSERVICE_ERROR =
```
janderson2, line 23: It appears this error message is only used for calls to unsupported methods. Perhaps a more explicit error message to this affect would make it more clear in the logs when debugging.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.warn;
+
+public class JWTPermissionsServiceImpl implements PermissionsService {
+    static final String PUBLISHER_PERMISSIONS = "publisher";
+    static final String ADMIN_PERMISSIONS = "admin";
```
janderson2, line 22: These values will need to match the group names rather than the type of permissions. These are the default groups for admins and publishers which are [`role-admin`](https://github.com/ONSdigital/dp-identity-api/blob/develop/models/groups.go#L14) and [`role-publisher`](https://github.com/ONSdigital/dp-identity-api/blob/develop/models/groups.go#L17) respectively:

```suggestion
    static final String PUBLISHER_GROUP = "role-publisher";
    static final String ADMIN_GROUP = "role-admin";
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        throw new JWTVerificationException(JWTPERMISSIONSSERVICE_ERROR);
+    }
+
+
+    /**
+     * @param session
+     * @param permission
+     * @return
+     */
+    public boolean hasPermission(Session session, String permission) {
```
janderson2, line 261: The name of this method is a little misleading since it is actually checking if a session is in a group. Also, this method should probably be private.
Perhaps something like:

```suggestion
    private boolean isGroupMember(Session session, String group) {
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,355 @@
...
+
+    public JWTPermissionsServiceImpl(Sessions sessionService) {
+    }
+
+    /**
+     * @param session - {@link Sessions} to get the groups
+     * @return list of groups where groupId is converted to teamId
+     * @throws IOException if session has no groups or empty list
+     */
+    public static List<Integer> convertGroupsToTeams(Session session) throws IOException {
```
daiLlew, line 43: Need to check if session is null as well + unit test to cover that scenario.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,355 @@
...
+            throw new IOException("JWT Permissions service error for convertGroupsToTeams no groups ");
+        }
+        String[] groups = session.getGroups();
+        Set<String> setOfString = new HashSet<>(
+                Arrays.asList(groups));
+        List<Integer> teamsList = new ArrayList<>();
+        for (String s : setOfString) {
+            try {
+                teamsList.add(Integer.parseInt(s));
+            } catch (NumberFormatException ignored) {
```
daiLlew, line 54: Need to log an error here so we can see which team failed.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,355 @@
...
+    public boolean canView(Session session, CollectionDescription collectionDescription) throws IOException {
+        List<Integer> teams = convertGroupsToTeams(session);
+        try {
+            AccessMapping accessMapping = permissionsStore.getAccessMapping();
+            Set<Integer> collectionTeams = accessMapping.getCollections().get(collectionDescription.getId());
+            if (collectionTeams == null || collectionTeams.isEmpty()) {
+                return false;
+            }
+            return teams.stream().anyMatch(t -> collectionTeams.contains(t));
+        } catch (IOException ignored) {
```
daiLlew, line 234: If something goes wrong here its because there was an issue reading the accessing mapping file which IMO is something we want/need to be aware of.

Can you update this throw that exception please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+ * perform any permisisons logic it determines which instance to invoke (legacy or JWT) based on the value of the
+ * feature flag.
+ */
+public class PermissionsServiceProxy implements PermissionsService {
+
+    private final PermissionsService legacyPermissionsService;
+    private final PermissionsService jwtPermissionsService;
+    private final boolean jwtSessionsEnabled;
+
+    public PermissionsServiceProxy(boolean jwtSessionsEnabled,
```
daiLlew, line 28: javadoc

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * @param email   the email of the user to remove the permission from.
+     * @param session the {@link Session} of the user revoking the permission.
+     * @throws IOException
+     * @throws UnauthorizedException
+     */
+    @Override
+    public void removeAdministrator(String email, Session session) throws IOException, UnauthorizedException {
+        if (jwtSessionsEnabled) jwtPermissionsService.removeAdministrator(email, session);
```
daiLlew, line 127: Can you add in the `{}` for the if-statement please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * @param session the {@link Session} of the user to check.
+     * @return the required module on jwtSessionsEnabled
+     * @throws IOException
+     */
+    @Override
+    public boolean canEdit(Session session) throws IOException {
+        if (jwtSessionsEnabled) return jwtPermissionsService.canEdit(session);
```
daiLlew, line 138: Can you add in the `{}` for the if-statement please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * @param email the email of the user to check.
+     * @return the required module on jwtSessionsEnabled
+     * @throws IOException
+     */
+    @Override
+    public boolean canEdit(String email) throws IOException {
+        if (jwtSessionsEnabled) return jwtPermissionsService.canEdit(email);
```
daiLlew, line 150: Can you add in the `{}` for the if-statement please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * @param user the {@link User} to check.
+     * @return the required module on jwtSessionsEnabled
+     * @throws IOException
+     */
+    @Override
+    public boolean canEdit(User user) throws IOException {
+        if (jwtSessionsEnabled) return jwtPermissionsService.canEdit(user);
```
daiLlew, line 162: Can you add in the `{}` for the if-statement please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+     * @param session the {@link Session} of the {@link User} granting the permissison.
+     * @throws IOException
+     * @throws UnauthorizedException
+     * @throws NotFoundException
+     * @throws BadRequestException
+     */
+    @Override
+    public void addEditor(String email, Session session)
+            throws IOException, UnauthorizedException, NotFoundException, BadRequestException {
+        if (jwtSessionsEnabled) jwtPermissionsService.addEditor(email, session);
```
daiLlew, line 178: Can you add in the `{}` for the if-statement please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceProxy.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * @param email   the email of the user to revoke the permission from.
+     * @param session the {@link Session} of the {@link User} revoking the permissison.
+     * @throws IOException
+     * @throws UnauthorizedException
+     */
+    @Override
+    public void removeEditor(String email, Session session) throws IOException, UnauthorizedException {
+        if (jwtSessionsEnabled) jwtPermissionsService.removeEditor(email, session);
```
daiLlew, line 191: Can you add in the `{}` for the if-statement please.

daiLlew (commented):
Functionally it looks fine just some minor things to tidy up.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,391 @@
...
+        if (session == null ||
+                collectionDescription == null ||
+                session.getGroups() == null ||
+                collectionDescription.getTeams() == null) {
+            throw new IOException("Empty or Null Session or CollectionDescription");
+        }
+
+        Set<Integer> resultTeamIds = new HashSet<>();
+        for (Integer sessionTeamID : convertGroupsToTeams(session)) {
+            for (Integer collectionTeamIds : convertTeamsListStringToListInteger(collectionDescription)) {
```
daiLlew, line 266: The logic looks correct but I think we can refactor this slightly:

```
HashSet<Integer> userPermissions = convertGroupsToTeams(session);
HashSet<Integer> collectionTeams = convertTeamsListStringToListInteger(collectionDescription);

List<Integer> result = userPermissions.stream()
                .filter(collectionPermissions::contains)

return result != null && !result.empty()
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImpl.java
```diff
@@ -0,0 +1,391 @@
...
+        Set<Integer> resultTeamIds = new HashSet<>();
+        for (Integer sessionTeamID : sessionsTeamsList) {
+            for (Integer collectionTeamIds : collectionTeamIntegerList) {
+                if (sessionTeamID.equals(collectionTeamIds)) {
+                    resultTeamIds.add(sessionTeamID);
+                }
+            }
+        }
+
+        return java.util.Collections.unmodifiableSet(resultTeamIds);
```
daiLlew, line 344: Im not 100% sure this logic is right for what we need to do? I think this needs to return a list of team IDs for the teams that have access to the collection. If that is correct then we don't need the session as this will be the user's permissions.

daiLlew (changes requested):

daiLlew (approved):

PR #553, opened 2021-12-08
Tidy up sessions handling prior to enabling JWT sessions feature

janderson2:
### What

Rework the sessions handling to prepare for the new JWT sessions being
enabled. Move the JWT session implementation to the service package
rather than store package since it implements the service interface.

This includes:
* Deprecating the old sessions logic ready for deletion after the migration is complete
* Moving the new sessions service implementation under the service package
* Write out the `Sessions.find(email)` logic as this functionality will not be possible when using the JWT sessions
* Write out the public `Sessions.expired(session)` logic as this was not actually adding value since all usages were actually called immediately after another method that had already checked for expiry

~🚨 Depends on https://github.com/ONSdigital/zebedee/pull/554 to resolve the audit failures~ (Merged)

### Who can review

Anyone but me.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/keyring/CollectionKeyringUtil.java
```diff
@@ -29,6 +29,8 @@ private CollectionKeyringUtil() {
      * @throws InternalServerError unexpected error getting user.
      * @throws NotFoundException   thrown if user not found.
      * @throws BadRequestException thrown of the user email is invalid.
+     *
+     * //TODO: Review usage of this method and see if it can be written out or updated to use the JWT session to get the user
```
janderson2, line 33: This TODO will been addressed in a subsequent PR.

janderson2:
Just changed to base branch back and forth because Github didn't update the diff against develop after the last merge.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Zebedee.java
```diff
@@ -16,11 +16,7 @@
 import com.github.onsdigital.zebedee.model.encryption.EncryptionKeyFactory;
 import com.github.onsdigital.zebedee.model.publishing.PublishedCollections;
 import com.github.onsdigital.zebedee.permissions.service.PermissionsService;
-import com.github.onsdigital.zebedee.service.DatasetService;
-import com.github.onsdigital.zebedee.service.ImageService;
-import com.github.onsdigital.zebedee.service.KafkaService;
-import com.github.onsdigital.zebedee.service.ServiceStore;
-import com.github.onsdigital.zebedee.service.ServiceStoreImpl;
+import com.github.onsdigital.zebedee.service.*;
```
daiLlew, line 19: Can you update your ide settings to not user wildcard imports please.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/exceptions/SessionsException.java
```diff
@@ -0,0 +1,11 @@
+package com.github.onsdigital.zebedee.session.service.exceptions;
+
+public class SessionsException extends RuntimeException {
```
daiLlew, line 3: Javadoc.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/exceptions/SessionsTokenExpiredException.java
```diff
@@ -1,6 +1,6 @@
-package com.github.onsdigital.zebedee.session.store.exceptions;
+package com.github.onsdigital.zebedee.session.service.exceptions;
```
daiLlew, line 1: Is this subclasses needed anymore? It doesn't offer anything the parents class doesn't, the only reason I can think of is that we use the exception class to determine specific error types. If that it not used then we could just delete this class.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/exceptions/SessionsVerificationException.java
```diff
@@ -1,6 +1,6 @@
-package com.github.onsdigital.zebedee.session.store.exceptions;
+package com.github.onsdigital.zebedee.session.service.exceptions;
```
daiLlew, line 1: Ditto

daiLlew (commented):
Logic seems fine but you need to update your IDE to disable wildcard imports. I've also raised a question about an exception subclass

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Zebedee.java
janderson2: 🙈 Thanks, have set up a new mac and haven't got the settings right again.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/exceptions/SessionsTokenExpiredException.java
janderson2: Removed.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/session/service/exceptions/SessionsVerificationException.java
janderson2: Removed this as well among a couple others to which this applies.

daiLlew (approved):

PR #567, opened 2021-12-29
Filter out invalid dataset uris for kafka event

bpathak-ons:
### What

* Do not trigger kafka event for invalid dataset uris
* Trello task - https://trello.com/c/e4ctBF6b/963-update-zebedee-publish-step-to-not-trigger-kafka-event-for-dataset-url

### How to review

* see if changes make sense
* CI pass

### Who can review

* Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
@@ -764,7 +764,7 @@ public class Publisher {
             List<String> datasetUris = collection.getDatasetVersionDetails()
                     .stream()
                     .map(content -> convertUriForEvent(content.uri))
-                    .filter (Publisher::isValidUris)
+                    .filter (Publisher::isValidCMDDatasetURI)
                     .collect(Collectors.toList());
 
             info().data("collectionId", collection.getId())
@@ -784,9 +784,10 @@ public class Publisher {
         sendMessage (collection, reviewedUris, "Reviewed-uris");
     }
 
-    // valid uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
-    protected static boolean isValidUris (String uri){
-        return uri.split("/").length > 6;
+    // Valid CMDDataset uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
+    protected static boolean isValidCMDDatasetURI (String uri){
+
+        return uri.chars().filter(ch -> ch == '/').count() > 6;
     }
 
     // Putting message on kafka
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -764,11 +764,19 @@ private static void sendToKafka(Collection collection) throws IOException {
             List<String> datasetUris = collection.getDatasetVersionDetails()
                     .stream().map(content -> convertUriForEvent(content.uri))
                     .collect(Collectors.toList());
+
+            // valid uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
+            List<String> validDatasetUris = new ArrayList<>();
```
jessjenkins, line 769: Rather than creating a new array, why not use the already existing stream above to filter out the unwanted URI's before the stream is collected to a list?
You can then have a separate validation function which can be cleanly unit tested. See section 2.1 in this tutorial for an example. https://www.baeldung.com/java-stream-filter-lambda

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -781,6 +784,11 @@ private static void sendToKafka(Collection collection) throws IOException {
         sendMessage (collection, reviewedUris, "Reviewed-uris");
     }
 
+    // valid uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
+    protected static boolean isValidUris (String uri){
```
jessjenkins, line 788: I'd rename this function as it's too broad. All the uris (or at least most) are presumably valid for something. The method name should indicate exactly what's being worked out. Eg. `isValidCMDDatasetURI` or something like that.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -781,6 +784,11 @@ private static void sendToKafka(Collection collection) throws IOException {
...
+        return uri.split("/").length > 6;
```
jessjenkins, line 789: I had some concerns about the (small) overhead of splitting the string and assigning them to a list just to count the occurrences. I looked for some guides o best ways to count a specific char in Java and there's an entire article on it…
https://www.baeldung.com/java-count-chars
Not really bothered as it's only a small operation so I'll still approve anyway but thought I'd give you the opportunity to decide for yourself.

bpathak-ons: Thanks for pointing out, method name now updated with readable code using core library

jessjenkins (approved):

PR #568, opened 2022-01-05
fix to capture version of changed dataset content type

bpathak-ons:
### What

* Fix to correctly capture uris for specific versions of dataset editions changes

### How to review

* look at changes and code works
* CI pass

### Who can review

* anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
@@ -787,7 +787,7 @@ public class Publisher {
     // Valid CMDDataset uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
     protected static boolean isValidCMDDatasetURI (String uri){
 
-        return uri.chars().filter(ch -> ch == '/').count() > 6;
+        return uri.chars().filter(ch -> ch == '/').count() > 5;
     }
 
     // Putting message on kafka
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
@@ -50,6 +50,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.locks.Lock;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.github.onsdigital.zebedee.api.Root.zebedee;
@@ -787,7 +789,15 @@ public class Publisher {
     // Valid CMDDataset uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
     protected static boolean isValidCMDDatasetURI (String uri){
 
-        return uri.chars().filter(ch -> ch == '/').count() > 5;
+        int matches = 0;
+        Matcher matcher = Pattern.compile("/").matcher(uri);
+        while(matcher.find()) {
+            matches++;
+        }
+        if (matches>5)
+            return true;
+
+        return false;
     }
 
     // Putting message on kafka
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -787,7 +787,7 @@ private static void sendToKafka(Collection collection) throws IOException {
     // Valid CMDDataset uris for published CMD versions of a dataset (edition) - /dataset/{datatsetId}/editions/{edition}/versions/{version}/metadata
     protected static boolean isValidCMDDatasetURI (String uri){
 
-        return uri.chars().filter(ch -> ch == '/').count() > 6;
+        return uri.chars().filter(ch -> ch == '/').count() > 5;
```
daiLlew, line 790: I think this is potentially too open to be able accurately determined if a value is a CMD uri. Also not sure of the specific context here but technically a `/datasets` is a valid CMD uri.

From the example in the test class I think this is checking a dataset version uri? If so I think a regex would be a better approach as there is a specific uri pattern for this:

`/datasets/{datasetID}/editions/{edition}/versions/{version}`

I think it should be reasonably straight forward to implement this checking if the input string starts with that pattern. I think this would be a more robust solution.

daiLlew (changes requested):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
bpathak-ons: I found this as simple and single line change with cleaner and more readable solution using core library. I have not tried but regex uses Pattern and Matcher possibly involving more line of code changes.

daiLlew: I appreciate you're updating the code that already existed but for the reasons stated above this should probably change as IMO it doesn't provide a rigorous enough validation to correctly reject values that don't meet the criteria which I why I think while we are here we should update it.

In terms of scope creep - if this were multiple methods/classes then we'd have to make judgement on cost vs return but in this case it's 1 method so the impact will be limited and it shouldn't take too long to implement.

daiLlew (changes requested):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
bpathak-ons: Done, regex pattern applied

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -787,7 +789,15 @@ private static void sendToKafka(Collection collection) throws IOException {
...
-        return uri.chars().filter(ch -> ch == '/').count() > 6;
+        int matches = 0;
+        Matcher matcher = Pattern.compile("/").matcher(uri);
+        while(matcher.find()) {
+            matches++;
+        }
+        if (matches>5)
+            return true;
+
+        return false;
```
daiLlew, line 800: This is suffers from the same problem - the check is not specific enough to accurately identify a CMD dataset uri. For example if the input `uri` value is `/1/2/3/4/5` then that will be a match even though its not a valid CMD version uri.

You're correct in that the original code is a more concise implementation of this, but I was trying to say we should use a more accurate method to match the uris  which is why I suggested a regex - sorry I should have been more clear.

I suggested a regex because a valid CMD version uri has a specific pattern we can match against:

`/datasets/{dataset-id}/editions/{edition}/versions/{version}`

If we use a regex to match this pattern we can be way more accurate in our validation. I'm not sure this example is 100% correct & may need some refining but something to this effect should be what we need:

`^\/datasets\/.+\/editions\/.+\/versions\/[0-9]+$`

If we compile this as a regex we should be able to more accurately match values. I use this [online tool](https://www.regexpal.com/) to help me build up the regex I need in isolation which is pretty useful as it means you don't need to run the code to test it.

daiLlew (changes requested):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
bpathak-ons: Done, new regex pattern applied, pls let me know the specific expected regex, if not, which meets the requirements.

daiLlew (approved):

PR #569, opened 2022-01-05
refactored CMS start up - it no longer uses falls back on default con…

daiLlew:
- Refactored CMS start up so it no longer falls back on default content if there is an error starting up. 
  
  Previously any error starting up would be caught and it would attempt to start up again using the default content. This functionality is misleading, masks issues and makes the CMS seem healthy when its not. 

  With these new changes any error initialising the CMS will causing the application to shutdown with an explicit message about the error encountered. 
- Updated `Restolino` to latest version.
- Added `@Priority(n)` annotations to init classes to guarantee start up order - Reader 1st, CMS 2nd.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
@@ -1,5 +1,6 @@
 package com.github.onsdigital.zebedee;
 
+import com.github.davidcarboni.restolino.framework.Priority;
 import com.github.davidcarboni.restolino.framework.Startup;
 import com.github.onsdigital.logging.v2.DPLogger;
 import com.github.onsdigital.logging.v2.Logger;
@@ -22,6 +23,7 @@ import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
 /**
  * Created by bren on 31/07/15.
  */
+@Priority(2)
 public class Init implements Startup {
 
     private static final String FORMAT_LOGS_KEY = "FORMAT_LOGGING";
@@ -42,6 +44,8 @@ public class Init implements Startup {
             System.err.println(ex);
             System.exit(1);
         }
+
+        info().log("initialising Zebedee CMS");
 
         info().log("loading CMS feature flags");
         CMSFeatureFlags.cmsFeatureFlags();
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
@@ -16,7 +16,8 @@ import com.github.onsdigital.zebedee.model.ZebedeeCollectionReaderFactory;
 import com.github.onsdigital.zebedee.persistence.dao.CollectionHistoryDaoFactory;
 import com.github.onsdigital.zebedee.reader.ZebedeeReader;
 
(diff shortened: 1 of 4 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
```diff
@@ -126,15 +85,61 @@ public static void init() {
...
 
+    /**
+     * Validate the zebedee root path meets the requirements - not null/empty, the file exists and is a directory.
+     * Throws {@link RuntimeException} if any of the checks fail.
+     *
+     * @param rootDir the path to validate.
+     * @return the root dir value as a {@link Path} object.
+     */
+    private static Path validateZebedeeRootConfig(String rootDir) {
+        if (StringUtils.isEmpty(rootDir)) {
```
s-xton, line 103: This check does not seem to get called, see 

ReaderConfiguration line 105

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
```diff
@@ -126,15 +85,61 @@ public static void init() {
...
+
+        if (!Files.isDirectory(p)) {
+            throw new RuntimeException(
+                    format("invalid zebedee_root config:{0} the specified path is not a directory", rootDir));
+        }
+
+        info().data(ZEBEDEE_ROOT, rootDir)
+                .log("successfully validated zebedee_root config proceeding beginning CMS initialization");
+        return p;
+    }
```
s-xton, line 121: just a thought, I was wondering if you check for one or more of the known directories here as Zebedee will start up if the env variable to set to an empty directory ?

s-xton (changes requested):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Root.java
daiLlew: ReaderConfiguration line 153 assigns it to the class field which is then used in the subsequent constructor logic.

s-xton (dismissed):

s-xton:
Tested with empty zedebee_root
Tested with zedebee_root to invalid location
Tested with zedebee_root to a file location
Tested with zedebee_root to a directory location

commented issue is down to race condition in existing framework

Marianasilvago (approved):
LGTM

Marianasilvago (approved):
LGTM

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/Init.java
```diff
@@ -16,11 +17,13 @@
...
 import com.github.onsdigital.zebedee.reader.ZebedeeReader;
 
-import static com.github.onsdigital.logging.v2.event.SimpleEvent.info;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.error;
+import static com.github.onsdigital.zebedee.logging.CMSLogEvent.info;
 
 /**
  * Created by bren on 31/07/15.
  */
+@Priority(2)
```
daiLlew, line 26: Annotation specifies the order the `Startup` instances should run. The app logging is configured within Reader so we want reader to start before CMS so we don't miss any log events.

franmoore05 (approved):
LGTM

PR #570, opened 2022-01-06
remove the default content dir - no longer required

daiLlew:
Removing the default content as it's not longer required. **No code changes**, only deleted the content dir which content example data.json files.

Depends on - https://github.com/ONSdigital/zebedee/pull/569


mike-kelly-methods (approved):
👍

PR #573, opened 2022-01-13
Rename enum values to follow convention

rafahop:
### What

Enums values are not currently following naming convention: they should be uppercase characters separated by underscores. Presumably this is so that it serialises to the value we expect (lowercase). 

Rename the enum values to follow conventions and add the `@SerializedName` annotation so that they serialise to the required (lowercase) string.

Also took the opportunity to make `ContentDetail` use a `PageType` enum rather than its string representation. This makes the class more robust by only accepting valid types and helps ensuring this is not a breaking change. 

### How to review

Check changes make sense. 
Test creating/reading content still works (check page type and alert types)
Test elastic search interaction still works

### Who can review

Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/ContentDetail.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/ContentDetail.java
@@ -5,12 +5,14 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.github.onsdigital.zebedee.content.page.base.PageType;
+
 /**
  * Class to hold a file uri and any other properties required from that file.
  */
 public class ContentDetail {
     public String uri;
-    public String type;
+    private PageType type;
     public ContentDetailDescription description;
 
     public List<ContentDetail> children;
@@ -20,6 +22,17 @@ public class ContentDetail {
 
     public ContentDetail() {
     }
+    
+    /**
+     * Convenience constructor
+     *
+     * @param uri
+     * @param type
+     */
+    public ContentDetail(String uri, PageType type) {
+        this.uri = uri;
+        this.type = type;
+    }
 
     /**
      * Convenience constructor taking the typical parameters.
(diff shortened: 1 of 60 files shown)


Marianasilvago (dismissed):
LGTM

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/partial/AlertType.java
```diff
@@ -1,6 +1,10 @@
 package com.github.onsdigital.zebedee.content.partial;
 
+import com.google.gson.annotations.SerializedName;
+
 public enum AlertType {
-    alert,
-    correction,
+    @SerializedName("alert")
+    ALERT,
```
daiLlew, line 7: Can you put a new line between these please.

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/partial/AlertType.java
```diff
@@ -1,6 +1,10 @@
...
+    @SerializedName("correction")
```
daiLlew, line 8: Do you also want to add the get label functionality to this type as well?

rafahop: There is no need for that get label method here as this is only used for serialising/deserialising. I had to add the method in the page type because it is widely used somewhere else and needed to have a way of retrieving the value

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/ContentDetail.java
```diff
@@ -20,6 +22,17 @@
 
     public ContentDetail() {
     }
+    
+    /**
+     * Convenience constructor
+     *
+     * @param uri
+     * @param type
+     */
```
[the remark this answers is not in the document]
rafahop: I was just trying to be consistent with the other constructors but you are right. Will remove it

daiLlew (approved):

PR #580, opened 2022-01-18
fixed URI builder in pooled http client

daiLlew:
### Fix mathjax equation request builder.

We recently upgraded our version of `org.apache.httpcomponents.httpclient` from `4.5.1` to `4.5.13`. The default `URIBuilder.path` value in the old version returned `""` whereas the latest version returns `null`. This results in the `PooledHttpClient.uriBuilder` building requests with an incorrect URI:

After upgrading the URI builder is returning:
```
http://{host}:{port}/null/
```
instead of 
```
http://{host}:{port}/
```
Which results in a request to an invalid/unknown endpoint in our Mathjax server. I've updated our code to default a `null` path  value to `""` instead of null to maintain the original functionality.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/verification/http/PooledHttpClient.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/verification/http/PooledHttpClient.java
@@ -164,7 +164,8 @@ public class PooledHttpClient {
 
     private URIBuilder newUriBuilder(String path) {
         URIBuilder uriBuilder = new URIBuilder(HOST);
-        uriBuilder.setPath((uriBuilder.getPath() + "/" + path).replaceAll("//+", "/"));
+        String uri = StringUtils.defaultIfEmpty(uriBuilder.getPath(), "");
+        uriBuilder.setPath((uri + "/" + path).replaceAll("//+", "/"));
         return uriBuilder;
     }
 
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/verification/http/PooledHttpClient.java
```diff
@@ -164,7 +164,8 @@ private URI buildPath(String path) {
 
     private URIBuilder newUriBuilder(String path) {
         URIBuilder uriBuilder = new URIBuilder(HOST);
-        uriBuilder.setPath((uriBuilder.getPath() + "/" + path).replaceAll("//+", "/"));
+        String uri = StringUtils.defaultIfEmpty(uriBuilder.getPath(), "");
```
daiLlew, line 167: If `uriBuilder.getPath()` returns `null` then `uriBuilder.getPath() + "/"` results in `"null/"`.

justinpjose (approved):
Tested it locally and it works. Awesome work Dave

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/verification/http/PooledHttpClient.java
goofballLogic: looks good to me but do you want a more qualified eye to look at it?

PR #582, opened 2022-02-02
added new collection event for attach cal entry and delete collection…

daiLlew:
### What

There has been several Calendar related issues on support recently so added some additional info to the collection event history to help diagnose code errors vs human error.

- Added new collection `Event` - `CALENDAR_ENTRY_ATTACHED_ON_CREATION`.
- Updated collection creation code to add above event when a new collection is created if its attached to a cal entry.
- Updated delete collection content to add `DELETED` event to the collection events by uri array.
- Updated tests.

The image below shows the collection JSON with these changes in effect.

1. Events by URI shows contains an entry for the `release` page.
2. An Event by URI shows that the release page was deleted from the collection.
3. A collection event stating that the collection was created attached to a calendar entry (also includes the name & uri of the entry for reference).

Screenshot 2022-02-02 at 10 55 56

The hope is these additional events in the collection history will enable us to more easily/accurately reverse engineer the state of the collection at a given time to determined what may have happened & why.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/CollectionDescription.java
```diff
@@ -93,13 +95,37 @@ public CollectionDescription(String name, Date publishDate) {
...
+    }
+
+    /**
+     * Add an event to the collection history.
+     *
+     * @param date      the date of the event.
+     * @param eventType the type of event to record.
+     * @param session   the session of the user who triggered the event.
+     */
+    public void addEvent(Date date, EventType eventType, Session session, String note) {
```
nshumoogum, line 115: Should we be adding the note param to the documentation above?

nshumoogum (changes requested):
Looks fine, just 1 comment

PR #585, opened 2022-02-08
Add local dev auto config for JWT keys

janderson2:
### What

Add means of automatically setting the JWT verification key variables to
make it easier for developers to get up and running locally.

### How to review

Ensure that running the script locally with the `ENABLE_JWT_SESSIONS` env var set to true correctly starts zebedee without errors.

### Who can review

Anyone but me.


janderson2:
This solution won't work. The keys are in the incorrect format. Will find a better way.

PR #587, opened 2022-02-09
Refactor ZebedeeConfiguration for readbility

janderson2:
### What

Structure ZebedeeConfiguration in such a way as to keep the
initialisation of each service together and to ensure that dependencies
between services are clear.

This is the first in a series of PRs to attempt to organise the startup to be easier to follow.

### How to review

Ensure the changes make sense.

### Who can review

Anyone but me.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -151,301 +148,434 @@ public ZebedeeConfiguration(Path rootPath, boolean enableVerificationAgent) thro
...
+    /**
+     * Initialise the permissions service.
+     *
+     * @param zebedeePath the zebedee base directory (i.e. `$zebedee_root/zebedee`)
+     * @param teamsService the {@link TeamsService}
+     * @throws IOException If a filesystem error occurs.
+     */
+    private void initPermissionsService(Path zebedeePath, TeamsService teamsService) throws IOException {
+        Path permissionsPath = createDir(zebedeePath, PERMISSIONS);
         initialisePermissions(permissionsPath);
```
daiLlew, line 315: I know this is how you found it but can we rename this to something like `initialiseAccessMapping(permissionsPath)` to make it clearer what this does and to differentiate it from the method we're already in.

daiLlew (commented):
I've reviewed the changes and they make sense to me. I've checkout the branch and run the unit tests and integration test suite we have. I've also run it in my local stack and created and published a collection & everything appears to be working correctly.

I've added 1 minor comment otherwise Im happy to approve.

PR #588, opened 2022-02-11
Store topics

rafahop:
### What

Accept primary topic and secondary topics
Remove unused fields,  rename methods to follow standards and remove ternary operator.

### How to review

Check sense.

Optionally, it can be tested by posting content with `primaryTopic` and `secondaryTopics` under `description` and then retrieving it, checking that the values have been stored

Additionally, it can be tested using the new tags widget in Florence - see https://github.com/ONSdigital/florence/pull/673

### Who can review

Anyone

### DIFF
--- a/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/base/PageDescription.java
+++ b/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/base/PageDescription.java
@@ -81,8 +81,8 @@ public class PageDescription extends Content implements Comparable<PageDescripti
     private String versionLabel;
     private String metaCmd;
     private String apiDatasetId;
-    private Integer primaryTopic;
-    private List<Integer> secondaryTopics;
+    private String primaryTopic;
+    private List<String> secondaryTopics;
 
     public PageDescription() {
     }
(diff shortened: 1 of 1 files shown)


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/base/PageDescription.java
```diff
@@ -128,9 +128,8 @@ public void setMetaDescription(String metaDescription) {
         this.metaDescription = metaDescription;
     }
 
-
     public boolean isNationalStatistic() {
-        return nationalStatistic == null ? false : nationalStatistic;
+        return nationalStatistic != null && nationalStatistic;
```
daiLlew, line 132: I think you've already done so but can you make sure you update all methods in this class of this type to follow the same pattern you've applied here. Any small consistency wins we can get through is plus for me.

daiLlew (commented):
You code is fine just want you check you've updated all instances of the boolean method.

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/page/base/PageDescription.java
rafahop: Thanks @daiLlew! I thought I did but just double checked and yes, all boolean getters in this class follow this pattern now 🙂

daiLlew (approved):

PR #593, opened 2022-02-23
Feature/update collection remove teams logic

AWitcherONS:
### What

Update Team and Collections id 

### How to review
1) Java Code complies with current standards 
2) the maven package build is successful (this will include all unit tests)
3) dp-zebedee-keyring-migration-tests 3 will fail due to int /string  json marshal
solution 
modify dp-zebedee-sdk-go models 
type Team struct {
	**ID      string   `json:"id"`**
	Name    string   `json:"name"`
	Members []string `json:"members"`
modify dp-zebedee-keyring-migration-tests go.mod
replace github.com/ONSdigital/dp-zebedee-sdk-go => ../../dp-zebedee-sdk-go

### Who can review

anyone but me


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -73,28 +73,9 @@
...
-import static com.github.onsdigital.zebedee.Zebedee.KEYRING;
-import static com.github.onsdigital.zebedee.Zebedee.PERMISSIONS;
-import static com.github.onsdigital.zebedee.Zebedee.PUBLISHED;
-import static com.github.onsdigital.zebedee.Zebedee.PUBLISHED_COLLECTIONS;
-import static com.github.onsdigital.zebedee.Zebedee.SERVICES;
-import static com.github.onsdigital.zebedee.Zebedee.SESSIONS;
-import static com.github.onsdigital.zebedee.Zebedee.TEAMS;
-import static com.github.onsdigital.zebedee.Zebedee.USERS;
-import static com.github.onsdigital.zebedee.Zebedee.ZEBEDEE;
+import static com.github.onsdigital.zebedee.Zebedee.*;
```
daiLlew, line 76: Can you update your wildcard import settings please - We should specifically importing the packages/classes we need rather than everything.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -73,28 +73,9 @@
...
-import static com.github.onsdigital.zebedee.configuration.Configuration.getDatasetAPIURL;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getImageAPIURL;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getKafkaContentPublishedTopic;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getKafkaURL;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getKeyringInitVector;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getKeyringSecretKey;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getServiceAuthToken;
-import static com.github.onsdigital.zebedee.configuration.Configuration.getSlackSupportChannelID;
-import static com.github.onsdigital.zebedee.configuration.Configuration.slackChannelsToNotfiyOnStartUp;
+import static com.github.onsdigital.zebedee.configuration.Configuration.*;
```
daiLlew, line 78: Same here.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/ZebedeeConfiguration.java
```diff
@@ -103,34 +84,34 @@
...
+    private final Content published;
+    private final PermissionsService permissionsService;
+    private final UsersService usersService;
+    private final TeamsService teamsService;
+    private final Sessions sessions;
+    private final DataIndex dataIndex;
+    private final PermissionsStore permissionsStore;
+    private final DatasetService datasetService;
+    private final ImageService imageService;
+    private final KafkaService kafkaService;
```
daiLlew, line 111: These don't need to be declared `final`, a final modifier indicates that the reference shouldn't be updated but I am not sure that is a requirement or behaviour we want enforce here.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -39,12 +39,12 @@
...
-    private com.github.onsdigital.zebedee.model.Collections collections;
-    private UsersService usersService;
-    private CollectionKeyring collectionKeyring;
-    private ScheduleCanceller scheduleCanceller;
+    private final Sessions sessionsService;
+    private final PermissionsService permissionsService;
+    private final com.github.onsdigital.zebedee.model.Collections collections;
+    private final UsersService usersService;
+    private final CollectionKeyring collectionKeyring;
+    private final ScheduleCanceller scheduleCanceller;
```
daiLlew, line 47: Again final not required here.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collections.java
```diff
@@ -67,6 +67,30 @@ public Collections(ZebedeeCmsService zebedeeCmsService, DatasetService datasetSe
...
+     * @return
+     * @throws IOException
+     */
+    public static Collection getCollection(HttpServletRequest request)
+            throws IOException {
+        String collectionId = getCollectionId(request);
+        return Root.zebedee.getCollections().getCollection(collectionId);
+    }
+
+    public static String getCollectionId(HttpServletRequest request) {
```
daiLlew, line 83: Javadoc.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collections.java
```diff
@@ -34,19 +34,19 @@
...
 @Api
 public class Collections {
 
-    private ZebedeeCmsService zebedeeCmsService;
-    private DatasetService datasetService;
-    private final boolean datasetImportEnabled;
     private static final String GET_COLLECTIONS_ERROR = "get collections endpoint: unexpected error while attempting to get collections";
+    private final boolean datasetImportEnabled;
+    private final ZebedeeCmsService zebedeeCmsService;
+    private final DatasetService datasetService;
```
daiLlew, line 49: Final not required.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/DeletedContent.java
```diff
@@ -1,96 +0,0 @@
...
-
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
-import java.io.IOException;
-import java.util.List;
-
-
-@Api
```
daiLlew, line 26: Just checking why we are deleting this class? I think this is still in use

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Password.java
```diff
@@ -32,11 +32,11 @@
      * Wrap static method calls to obtain service in function makes testing easier - class member can be
      * replaced with a mocked giving control of desired behaviour.
      */
-    private ServiceSupplier<UsersService> usersServiceSupplier = () -> Root.zebedee.getUsersService();
+    private final ServiceSupplier<UsersService> usersServiceSupplier = () -> Root.zebedee.getUsersService();
```
daiLlew, line 35: final not required.

daiLlew (changes requested):

janderson2:
Not sure what has happened, but something seems to have gone wrong with the git history here. @AWitcherONS I am going to rebase this off develop again.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -130,10 +132,28 @@ public CollectionDescription get(HttpServletRequest request, HttpServletResponse
         return result;
     }
 
+    private void requireViewPermission(Session session, String collectionId) throws UnauthorizedException {
+        boolean canView = false;
+        try {
+            canView = permissionsService.canView(session, collectionId);
+        } catch (IOException ex) {
+            throw new UnauthorizedException("You are not authorised to view this collection");
```
daiLlew, line 140: This exception is thrown if there is an error checking the permission but we are swallowing the original exception. Can you update the message to reflect this, include the user email and collection ID and also include `ex` as a param. Something like:
```
throw new UnauthorizedException(MessageFormat.format("error checking user {0} view permissions for collection {1}", session.getEmail(), collectionID), ex);
```
You may need to add a new constructor to `UnauthorizedException` to take the additional param but I think its important to capture the original error details.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -130,10 +132,28 @@ public CollectionDescription get(HttpServletRequest request, HttpServletResponse
...
+        }
+
+        if (!canView) {
+            warn().user(session.getEmail())
+                    .collectionID(collectionId)
+                    .log("request unsuccessful user denied view permission");
```
daiLlew, line 146: I think we need to be smarter around what we log in Zebedee as its quite noisy. At present this will log the warning and eventually the exception stack trace. We don't need both so could we remove the warning log and update the exception below to include the user email and collection details. Something like:

```
throw new UnauthorizedException(MessageFormat.format("User is {0} not authorised to view collection {1}", session.getEmail, collectionID)
```
We get all the details we required but only once.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -202,6 +222,22 @@ public CollectionDescription create(HttpServletRequest request, HttpServletRespo
         return collection.getDescription();
     }
 
+    private void requireEditPermission(Session session) throws UnauthorizedException {
+        boolean canEdit = false;
+        try {
+            canEdit = permissionsService.canEdit(session);
+        } catch (IOException ex) {
+            throw new UnauthorizedException("You are not authorised to edit collections.");
```
daiLlew, line 230: Same here - update the message to reflect the broader error, include the session ID and the original exception in the constructor params.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -202,6 +222,22 @@ public CollectionDescription create(HttpServletRequest request, HttpServletRespo
...
+        }
+
+        if (!canEdit) {
+            warn().user(session.getEmail())
+                    .log("request unsuccessful user denied edit permission");
+
+            throw new UnauthorizedException("You are not authorised to edit collections.");
```
daiLlew, line 237: Same here - remove the warning log and update the exception to include all the necessary details to understand the context of this error.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
daiLlew: I know this code existed before but I'd like to take advantage and improve it while we're here.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/permissions/service/PermissionsServiceImplTest.java
```diff
@@ -31,13 +30,14 @@
...
+@SuppressWarnings("ALL")
 public class PermissionsServiceImplTest {
 
     private static final String EMAIL = "admin@ons.gov.uk";
     private static final String COLLECTION_ID = "123";
 
-    private Set<String> digitalPublishingTeam = new HashSet<>();
-    private Set<String> admins = new HashSet<>();
+    private final Set<String> digitalPublishingTeam = new HashSet<>();
+    private final Set<String> admins = new HashSet<>();
```
daiLlew, line 40: These don't need to be final/

daiLlew (changes requested):
There appears to be several large changes to code not related to teams in this PR. I know Zebedee has been refactored a lot recently so I am wondering if this is the result of a merge conflict/auto merge?

Im block this PR because I am worried we might be accidentally merging breaking changes. I appreciate this impacts you but I want to make sure we're not introducing a regression.

We can pair on this to resolve if that is helpful.

PR #594, opened 2022-02-23
Fix deletion of homepage from collection

rafahop:
### What

Fix issue when trying to delete homepages from a collection.  The uri to delete becomes `//data.json` which can't be found and the deletion fails. Fixing it by removing all trailing forward slashes from the received uri.

https://trello.com/c/f9XbZerw/374-cant-delete-homepage-from-collection-s2

### How to review

Check code and unit test.
Run florence, create a collection, edit the the homepage, save and delete it. It should disappear from the collection (and the filesystem) even when you go to another collection and come back to it.

### Who can review

Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Page.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Page.java
@@ -291,12 +291,7 @@ public class Page {
         }
 
         try {
-            String uriToDelete = uri;
-            // Remove trailing slashes
-            while (uriToDelete.endsWith("/")) {
-                uriToDelete = uriToDelete.substring(0, uriToDelete.length() - 1);
-            }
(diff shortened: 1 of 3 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Page.java
```diff
@@ -291,9 +291,16 @@ public void deletePage(HttpServletRequest request, HttpServletResponse response)
         }
 
         try {
+            String uriToDelete = uri;
+            // Remove trailing slashes
+            while (uriToDelete.endsWith("/")) {
+                uriToDelete = uriToDelete.substring(0, uriToDelete.length() - 1);
+            }
```
janderson2, line 298: This could potentially be simplified as:
```suggestion
           String uriToDelete = uri.replaceAll("/+$","");
```

janderson2 (approved):

PR #595, opened 2022-02-24
Fix content indexing

rafahop:
### What

As a result of https://github.com/ONSdigital/zebedee/pull/573, elastic search is now using the wrong indexes.
It should be using the label (serialised value) of the type rather than the enum name. All other instances were updated but this was missed 

https://trello.com/c/0NdPJkkW/782-zebedee-search-mapping-not-matching-elastic-search-templates

### How to review

Check the code makes sense

### Who can review

Anyone


daiLlew (approved):
I think this should fix the issue.

PR #597, opened 2022-03-03
Move populateQuietly to Collection object

rafahop:
### What

- Move the static `populateQuietly` method from `ReleasePopulator` to an instance method in `Collection`
Living unnecessarily in the `ReleasePopulator`, it has now been moved to `Collection`
and made the existing `populate` method private as not used anywhere else.
There is no need to keep both but left the non-quiet version just in case it
could be useful in the future and also to help with code review

- Extend unit tests to cover exceptional cases
- Change use of deprecated `FileUtils.write` method to include Charset

### How to review

Check changes make sense

### Who can review

Anyone


PatrickConnollyONS (approved):
Looks good to me, but if you want you could get another zebedee/java expert to have a look

PR #599, opened 2022-03-04
Fix log message

rafahop:
### What

Fix log messages: 
 - the creation hook wrongly mentions delete
 - the create endpoint wrongly mentions get

### How to review

Sense check

### Who can review

Anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Page.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Page.java
@@ -139,7 +139,7 @@ public class Page {
         try {
             session = zebedeeCmsService.getSession();
         } catch (ZebedeeException e) {
-            error().data("path", uri).logException(e, "page get endpoint: failed to get session");
+            error().data("path", uri).logException(e, "page create endpoint: failed to get session");
             response.setStatus(e.statusCode);
             return null;
         }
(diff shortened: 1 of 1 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Page.java
```diff
@@ -337,15 +337,15 @@ private boolean execCreationHook(com.github.onsdigital.zebedee.content.page.base
                                      Collection collection, Session session) {
         info().data("path", uri).data("collection_id", collection.getDescription().getId()).data("user",
                 session.getEmail())
-                .log("page delete endpoint: executing PageCreationHook");
+                .log("page create endpoint: executing PageCreationHook");
```
goofballLogic, line 340: how do you know which endpoint is calling this? maybe the "page create endpoint" bit of the string needs to be passed in?

rafahop: How do I know? 🤔  Just follow the trace and you'll see it is called from the create page endpoint

goofballLogic: shall i remove this comment and let someone else review instead?

mike-kelly-methods (approved):
👍

PR #605, opened 2022-03-09
Improve unit tests checking lapses of time

rafahop:
### What

These unit tests were failing randomly as they were checking times using a low tolerance. 

### How to review

Check changes make sense

### Who can review

Any dev


cookel2 (approved):
looks good to me

PR #617, opened 2022-04-12
update jwt-verifier version

Marianasilvago:
### What

Update jwt version to parse fetch jwt keys response.

### Who can review

Anyone.


mryan321 (approved):
😊

PR #620, opened 2022-04-22
Fix setting collection viewer teams (JWT enabled)

janderson2:
### What

A piece of technical debt involving a hack in zebedee that maps the team
names back to team IDs (because Florence is sending the names when it
shouldn't be) is causing creation of collections to fail with a 500 then
409. This is because the logic used to map the name to the ID is no
longer possible with the JWT sessions enabled and group management moved
to the dp-identity-api. Since Florence actually as the team ID to hand
the cleaner solution is to simply have florence provide the team IDs
rather than names in the first place. This change lays the ground work
for that change.

### How to review

Ensure that the team IDs submitted in the teams array will be stored as is when the JWT sessions are enabled, but that the old logic will still function the same way.

### Who can review

!me

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -441,7 +441,8 @@ public class Collection {
             for (String teamName : teamNames) {
                 Team team = teams.findTeam(teamName);
                 if (team == null) {
-                    throw new NotFoundException("team assigned to collection expected but does not exist");
+                    info().data("team", teamName).log("team assigned to collection expected but does not exist, skipping...");
+                    continue;
                 }
 
                 teamIds.add(team.getId());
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -424,24 +425,28 @@ public static Collection update(
...
+            if (teamNames == null) {
+                teamNames = new ArrayList<>();
             }
 
-            teamIds.add(team.getId());
+            TeamsService teams = zebedee.getTeamsService();
+            for (String teamName : teamNames) {
+                Team team = teams.findTeam(teamName);
+                if (team == null) {
+                    throw new NotFoundException("team assigned to collection expected but does not exist");
```
daiLlew, line 444: Need to be careful here as the team could be deleted before the collection is processed and the description updated. If that happens you will get this exception and the action you are trying to perform will fail.
I think it would be safer so quietly error here rather than block the action completing.

daiLlew (commented):
I've added 1 comment, but the unit tests do need to be updated.

PR #623, opened 2022-05-10
Store survey values

rafahop:
### What

Store survey values. Currently limited to `census` but open to extension via an enum

### How to review

Check changes make sense.

Optionally, it can be tested by posting content with `"survey" : "census"` in the description and then retrieving it, checking that it returns the survey value.

### Who can review

Anyone not me


nshumoogum (approved):
LGTM

PR #626, opened 2022-05-24
Extend to include management of interactives in collections

mryan321:
### What

Note: a squash/force to tidy and this PR also includes contributions from @nimitsarup 😊 

An `interactive` is a new type of "page" and this PR is to allow zebedee to handle the new type within a collections workflow similar to filterable datasets - from adding to a collection to publishing (this will hook into publish task and call interactives api to publish too).

Just for clarity - if you look in florence develop right now there is an `interactives` menu item - this will be dropped soon and the only entrypoint will be via a collection.

The florence UI (tbc) for interactives will make the relevant zebedee call for updating state - e.g.:

```
curl --location --request PUT 'http://localhost:8082/collections/collection2/interactives/28417c90-a3a2-40bc-bb16-2d408bbae58c' \
--header 'X-Florence-Token:' \
--header 'Content-Type: application/json' \
--data-raw '{
    "state": "Reviewed"
}'
```

Example of json collection:

```
"interactives": [
  {
    "id": "28417c90-a3a2-40bc-bb16-2d408bbae58c",
    "title": "title1",
    "state": "InProgress",
    "uri": "/interactives/lable-uPOPNJXf/embed",
    "lastEditedBy": "mryan321@gmail.com"
  },
  {
    "id": "40de0587-d6cf-4f3e-9827-4cc68a416ab7",
    "title": "title",
    "state": "InProgress",
    "uri": "/interactives/label-5vgbc4j9/embed",
    "lastEditedBy": "mryan321@gmail.com"
  }
]
```

### How to review

Sanity check - maybe pull and experiment

### Who can review

Ideally anyone with knowledge of zebedee

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -414,7 +414,6 @@
                             <artifactId>xmlbeans</artifactId>
                             <version>2.6.0</version>
                         </exlude>
-                        <!-- End -->
 
                         <!-- https://trello.com/c/xDqibumj -->
                         <exlude>
--- a/zebedee-cms/pom.xml
+++ b/zebedee-cms/pom.xml
@@ -270,6 +270,12 @@
             <artifactId>dp-static-files-api-client-java</artifactId>
             <version>1.1.1</version>
         </dependency>
+
+        <dependency>
+            <groupId>com.github.onsdigital</groupId>
+            <artifactId>dp-interactives-api-client-java</artifactId>
+            <version>0.2.0</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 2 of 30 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -284,6 +284,8 @@ public boolean deleteCollection(HttpServletRequest request, HttpServletResponse
             throw new NotFoundException("The collection you are trying to delete was not found");
         }
 
+        //soft-delete any interactives
```
jessjenkins, line 287: I'd just be careful not to commit this as-is. Could be confused as a description of the following `deleteCollection(…)` method call. Consider `TODO`

mryan321: oops - this was left-over accidentally - i didnt realise you can only delete a collection when its empty 🤠

nimitsarup (approved):

PR #628, opened 2022-05-31
Deprecate methods in PermissionService in preparation for dp-permissions-api integration

janderson2:
This PR was originally by @janderson2 in 2022 - I (@lindenmckenzie) have brought it up to date with the latest changes - the original PR text is below:

### What

Remove and deprecate the PermissionsService.isPublisher and PermissionsService.isAdministrator methods respectively in preparation for the migration to using the dp-permissions-api.

In order to support this migration remove unnecessary usages of these methods and remove the `POST /service` endpoint used for generating service tokens via the API. This endpoint is not used in the environments anyway as access permissions block it. The documentation here and in the operations docs have been updated accordingly.

### How to review

Ensure that the changes are valid.

### Who can review

!me


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -1020,7 +1020,7 @@ public boolean skipDatasetVersionsValidation(Long userKey, Long overrideKey, Ses
             return false;
         }
 
-        if (!permissionsService.isPublisher(session)) {
+        if (!permissionsService.canEdit(session)) {
             info().user(session)
                     .reason("publisher permissions required")
```
rafahop, line 1025: Would it make sense changing this log message?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/permissions/service/JWTPermissionsServiceImplTest.java
```diff
@@ -100,51 +99,59 @@ public void isPublisher_Session_NullGroup_ShouldReturnFalse() throws Exception {
...
 
     @Test
-    public void isAdministrator_Session_Publisher_ShouldReturnFalse() throws Exception {
+    public void isAdminUser_Session_Publisher_ShouldReturnFalse() throws Exception {
         List<String> sessionGroups = new ArrayList<>();
         sessionGroups.add(PUBLISHER);
 
         Session session = new Session(TEST_SESSION_ID, TEST_USER_EMAIL, sessionGroups);
-        assertFalse(jwtPermissionsService.isAdministrator(session));
+        assertFalse(jwtPermissionsService.isAdminUser(session));
```
rafahop, line 124: This is now a protected method and probably not worth keeping these tests here?

It seems `isAdminUser` is only present in `JWTPermissionsServiceImpl` so to be used by its child class `PermissionsServiceImpl` (from `isAdministrator`) which will go at some point. Maybe we could move it to the child class and these tests to a test class for `PermissionsServiceImpl` testing the public `isAdministrator` instead?

And do something similar for the `isPublisher` too.

rafahop (dismissed):

rafahop (approved):

janderson2 (commented):
LGTM. @lindenmckenzie over to you to review

lindenmckenzie (approved):

janderson2 (commented):
Still LGTM post rebase 👍

PR #629, opened 2022-06-01
Replaced the use of the spring-web dependency 

wbquinn:
Replaced the use of the spring-web dependency with a clone of the one class needed by spring-test

### What

The spring-web jar was added as a dependency to be able to use the sprint-test MockHttpRequest & MockHttpResponse objects. However there is an existing vulnerability in the spring-web library when using it to handle http connections so the audit process was failing.

This change removed the maven dependency and replicated the one required class (MediaType) needed for mocking the HttpResponse.

### How to review

Consider the approach of cloning a single class to avoid a vulnerability...

### Who can review

Any Java person


rafahop (changes requested):
I don't think we should be cloning Spring code? 
Please use Mockito if `MockHttpResponse` is not suitable for your tests. You can find an example of how this is done [here](https://github.com/ONSdigital/zebedee/blob/61095cad46af4eb8d66bb0e6de470e3de982b7bb/zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/TeamsReportTest.java#L102-L104)

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
```diff
@@ -143,17 +158,25 @@ public void writeVersionFileV1() throws Exception {
...
             e.printStackTrace();
             Assert.fail(e.getMessage() + e.toString());
             throw e;
         }
 
-        Assert.assertEquals(200, response.getStatus());
-        JSONAssert.assertEquals(EXPECTED_CONTENT_V1, response.getContentAsString(), false);
+        // admittedly a little fragile - ideally we'd like to be sure that the last invocation (or possibly default value) was correct.
+        verify(mockResponse, atLeast(1)).setStatus(HttpStatus.SC_OK);
+        verify(mockResponse, atLeast(1)).setContentType(MediaType.APPLICATION_JSON);
```
rafahop, line 177: Is using `atLeast` what makes this a little fragile? Should we use `times` instead?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
```diff
@@ -200,8 +217,9 @@ public void writeVersionFileV2() throws Exception {
             throw e;
         }
 
-        Assert.assertEquals(200, response.getStatus());
-        JSONAssert.assertEquals(EXPECTED_CONTENT_V2, response.getContentAsString(), false);
+        // Then I should receive a structured V1 content back with uri download link
```
rafahop, line 220: I still think we don't need 2 versions of this test but I suppose that it will be addressed in a separate PR (if addressed at all), for now this one should refer to V2
```suggestion
        // Then I should receive a structured V2 content back with uri download link
```

jon-acker: There are two separate behaviours of the system because it deserialises differently for each type of content, so there need to be two separate tests. 
I corrected the comment, thanks.

rafahop: I think I am missing something here as I can't see any difference in these tests. I'd appreciate if you or anyone in the team could explain better? Why would not [something like this](https://github.com/ONSdigital/zebedee/compare/wip/single-test) be acceptable?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
```diff
@@ -108,7 +116,7 @@ public void setUp() throws Exception {
     }
 
     @Test
-    public void writeVersionFileV1() throws Exception {
+    public void ResponseContainsFileDownloadLinkWhenJSONV1IsWritten() throws Exception {
```
rafahop, line 119: Still not following Java standards, it should start with lowercase
```suggestion
    public void responseContainsFileDownloadLinkWhenJSONV1IsWritten() throws Exception {
```

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
```diff
@@ -118,22 +126,21 @@ public void writeVersionFileV1() throws Exception {
...
-        // Create the version content via the content API.
+        // Given V1 Of the JSON was written
         try {
             Content contentApi = new Content();
             boolean result = contentApi.saveContent(request, new MockHttpServletResponse());
             Assert.assertTrue(result);
         } catch (Exception e) {
             e.printStackTrace();
-            Assert.fail(e.getMessage() + e.toString());
+            Assert.fail(e.getMessage() + e);
```
rafahop, line 136: This was actually also mentioned in the [original PR](https://github.com/ONSdigital/zebedee/pull/621#discussion_r868141374). I don't think we need this catch at all and we wouldn't need to be maintaining unnecessary code

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/StubServletOutputStream.java
```diff
@@ -0,0 +1,35 @@
+package com.github.onsdigital.zebedee.api;
```
rafahop, line 1: Maybe move it up to a different more generic package rather than make it specific to apis?

pom.xml
```diff
@@ -287,7 +287,7 @@
             <dependency>
                 <groupId>org.apache.tika</groupId>
                 <artifactId>tika-core</artifactId>
-                <version>1.27</version>
+                <version>1.28.3</version>
```
rafahop, line 290: This is fixed in develop now, please merge it back 🙂

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/StubServletOutputStream.java
jon-acker: it's only used by api tests

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
jon-acker: The test will not reporting a proper "failure message" if we leave it to throw an exception here, so was neccessary to catch.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/StubServletOutputStream.java
rafahop: I know, but we do not place classes in packages where they are used (at the moment) but where they belong? This is not api specific and could be used from anywhere else?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
rafahop: We do not expect an exception to occur. If it does, something has gone wrong and we should throw the exception. It is not a test failure but an error

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/StubServletOutputStream.java
jon-acker: what package do you suggest?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
jon-acker: sorry, my mistake I missed the other one...

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/StubServletOutputStream.java
jon-acker: I can't actually see a more suitable place, it's next to where it's used and unlikely to be used other than for API testing.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
braddle: That does represent what should happen.

The download will either have `file` or (`uri` and `version`) that suggested test represents an illegal state.

rafahop (changes requested):
It seems `Makefile.gpg` has been accidentally checked in, could it be deleted please?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ContentTest.java
braddle: But the two test represent that way that this part of the JSON should be use. V1 format (`file`) is the current way of working when the files are store and server in Zebedee. V2 format (`uri` and `version`) mean that the file is store and served by by the static files system (dp-upload, dp-download, dp-files).

The test should that both versions can work in parallel. The two test are showing two different behaviours of the same endpoint. 

If the data where in the illegal state (having all three fields set at the same time or having `version` and `files` set but not `url`) this would have the potential of breaking downstream systems of zebedee.

rafahop: Thanks, I understand that but it is not an illegal state as there are no rules checking it and hence the test with the 3 fields passes. The two tests do not show any different behaviour but only one: it can process the request by marshalling/unmarshalling the `DownloadSection`. I do not think there is any different logic in the endpoint when it receives a `version` and `files` than when it receives an `url`. It is dumb as far as I can see and we are only making sure it can accept the new fields. Having 2 tests doing the same thing (and not even sharing the code) is only creating technical debt and unnecessary maintenance.
But that is not really part of this pull request and should have been discussed and addressed earlier (see https://github.com/ONSdigital/zebedee/pull/621) instead of ignoring the review comments. On that basis, I don't think it should stop this branch from being merged 👍

braddle (approved):
LGTM - Pairing

PR #635, opened 2022-06-07
fix server error when getTeams returns null

gedge:
### What

when editing a collection (e.g. publishDate), zebedee 500's
this is a NullPointerException caused by getTeams returning null
this PR should avoid that

### How to review

Describe the steps required to test the changes.

### Who can review

Describe who worked on the changes, so that other people can review.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -412,9 +412,8 @@ public class Collection {
             scheduler.cancel(collection);
         }
 
-        Set<String> teamIds = setViewerTeams(collectionDescription, zebedee, session);
-
         if (collectionDescription.getTeams() != null) {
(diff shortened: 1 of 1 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -412,9 +412,8 @@ public static Collection update(
             scheduler.cancel(collection);
         }
 
-        Set<String> teamIds = setViewerTeams(collectionDescription, zebedee, session);
-
         if (collectionDescription.getTeams() != null) {
+            Set<String> teamIds = setViewerTeams(collectionDescription, zebedee, session);
```
mryan321, line 416: `teamIds` is not used 🤠  update the setTeams stmt to use or just delete?

gedge: the side-effect of `setViewerTeams` is the important thing here, but I agree that `teamIds` is not used (but I didn't change that aspect)

mryan321 (approved):

PR #640, opened 2022-06-10
Setting ENABLE_INTERACTIVES_PUBLISHING to "FALSE" for publisher testing

braddle:
### What

Bug Fix: A new test introduced the ENABLE_INTERACTIVES_PUBLISHING being set to true. This meant that the order test run in had the ability to affect the result of other tests.

### How to review

Ensure tests run locally usng IDE and make test-cms

### Who can review

Anyone


PauloPortugal (approved):
Cool!
Thanks for fixing it :)

jon-acker (approved):

PR #652, opened 2022-08-30
Feature/create array for cmd and cantabular datasets

cookel2:
# What has changed

The Release object has had a new attribute added to it so that cantabular and cmd datasets can be stored separately from other (legacy) datasets.

# How to review

The unit tests can be run using this command: make test

# Who can review

Anyone but me


cookel2:
> Looks good but I wonder if `RelatedApiDatasets` would be a suitable and more consistent name for this since the type refers to api dataset landing page?

Yes maybe. My understanding is that the DATASET_LANDING_PAGE is the old style timeseries pages, which can be left alone, whereas the API_DATASET_LANDING_PAGE corresponds to both the 'CMD' and 'Cantabular' types of release. My naming of relatedCantabularOrCMDDatasets is intended to help make this clear but I do get your point. However, I think the dp-release-calendar-api ( and backend aspect of dp-frontend-release-calendar) are being updated to cater for the new list and the new name might make more sense when used by those services. So I'll have a think.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
```diff
@@ -1397,6 +1397,50 @@ public void populateReleaseQuietlyShouldAddLinksToReleasePageForCollectionConten
...
+                collection.getDescription().getId());
+        Iterable<ContentDetail> collectionContent = ContentDetailUtil.resolveDetails(collection.getReviewed(),
+                collectionReader.getReviewed());
+
+        Release result = collection.populateReleaseQuietly(
+                collectionReader,
+                collectionWriter,
+                collectionContent);
+
+        // Then the release is now in progress for the collection and the published flag is set to true
```
rafahop, line 1437: This comment does not reflect what the code does?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
```diff
@@ -1397,6 +1397,50 @@ public void populateReleaseQuietlyShouldAddLinksToReleasePageForCollectionConten
...
+
+        collection.getDescription().setReleaseUri(uri);
+        collection.associateWithRelease(publisher1Session, release, collectionWriter);
+
+        String releaseJsonUri = uri + "/data.json";
+
+        collection.complete(publisher1Session, releaseJsonUri, recursive);
+        collection.review(publisher2Session, releaseJsonUri, recursive);
+
+        ContentDetail CMDDetail = new ContentDetail("My CMD dataset", "/some/uri", PageType.API_DATASET_LANDING_PAGE);
```
rafahop, line 1418: Can we rename this local variable to follow Java conventions and make it camelCase please?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
cookel2: Good point. I've changed it now.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
cookel2: yes - done

PR #653, opened 2022-09-06
Generate Welsh PDFs

rafahop:
### What

Generate the Welsh language version of any content in PDF if there is one.
Refactor `ContentLanguage` enum and tidy some code

### How to review

Create a collection in Florence containing a page with Welsh content.
Approve the collection
Check that a `page_cy.pdf` has been generated in the relevant collection folder, as well as the usual `page.pdf`

### Who can review

Anyone


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/base/ContentLanguage.java
```diff
@@ -0,0 +1,42 @@
+package com.github.onsdigital.zebedee.content.base;
+
+import java.util.stream.Stream;
+
+public enum ContentLanguage {
+    ENGLISH("en"), WELSH("cy");
```
jessjenkins, line 6: Just wondering if these constants should be `EN` and `CY`. Not objecting to them being what they are, just raising the thought. I'm happy to go with whatever you decide.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/approval/task/CollectionPdfGeneratorTest.java
```diff
@@ -40,27 +45,87 @@ public class CollectionPdfGeneratorTest {
...
+
+        // If a Welsh version is requested, it means there is no English version (as it would override it)
+        verify(contentReader, never()).getResource(uri+"/data_cy.json");
+        // Verify it generates the Welsh version only
+        verify(pdfService, times(1)).generatePdf(contentWriter, uri, ContentLanguage.WELSH);
+        verify(pdfService, never()).generatePdf(contentWriter, uri, ContentLanguage.ENGLISH);
+    }
+
+    @Test
+    public void shouldGeneratePdfForEnglishAndWelshContent() throws IOException, ZebedeeException {
```
jessjenkins, line 114: As per other comment, I don't think this is right. If we haven't modified the Welsh version, there's no need to regenerate it. PDF generation is computationally expensive so this could massively reduce approval times and potential issues for especially large collections.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/CollectionPdfGenerator.java
```diff
@@ -67,87 +48,60 @@ public CollectionPdfGenerator(PdfService pdfService) {
...
+                    r = contentReader.getResource(detail.uri + "/" + ContentLanguage.WELSH.getDataFileName());
+                } catch (Exception e) {
+                    // It's ok. If there is no Welsh data file, we won't generate a Welsh PDF
+                    r = null;
+                }
+
+                if (r != null) {
+                    generatePDFForContent(collection, contentWriter, detail.uri, ContentLanguage.WELSH.getId());
+                }
+            }
```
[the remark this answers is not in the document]
rafahop: Did you test it with a collection containing only Welsh changes? 
The problem is when the collection contains both English and Welsh changes. In that case we only receive one `ContentDetail` object here for that url, with the English info. This is because we use a `Set` [while finding them](https://github.com/ONSdigital/zebedee/blob/b72fad73eeaeee792d22effc05fca874c4891ff6/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/util/ContentDetailUtil.java#L31) and the [equals method](https://github.com/ONSdigital/zebedee/blob/b72fad73eeaeee792d22effc05fca874c4891ff6/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/json/ContentDetail.java#L144) for `ContentDetail` only takes into account the uri, not the language. I considered changing that but this is done deliberately and would break other functionality. 
If the Welsh version hasn't changed, I don't think we will regenerate the PDF as the check in line 69 would fail because there won't be a `data_cy.json` in the collection?

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/content/base/ContentLanguage.java
rafahop: I thought I'd rename it to use the full word for easier reading but it is only naming and can be changed back to `EN` and `CY` if preferred

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/CollectionPdfGenerator.java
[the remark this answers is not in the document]
jessjenkins: OK, so the 'correct' way to fix it would be to fix the issue with the sets (if content has a different language then by definition it can't be 'equal' to the other language content). However, as you say, this will have a lot of implications due to zebedee's poor language handling (I can see the content tree functionality as one example that relies on this 'bug'). This does need fixing at some point (and I believe will actually have benefits with respect to other welsh language bugs) but is probably too big in scope for this change. Can I suggest instead we add a `FIXME` to this block explaining why it's needed with references to the `ContentDetailUtil` and the `ContentDetail::equals(…)`/`hashCode()` methods as you've detailed above. It's OK to do this as a workaround but we really need to clarify that that is what it is.

PR #659, opened 2022-10-05
Fix bug where Welsh content was overwritten by English changes

rafahop:
### What

If English changes are present in a collection and we try to get the equivalent Welsh content, it should not retrieve the English changes but the published Welsh content.

Override the `resolveDataFilePath` method when working with collections so that it won't find the English version in there but go and find the published content [further down the line](https://github.com/ONSdigital/zebedee/blob/4a1c3797ad13c6b17ab2768914eaa8b3d8e025f3/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/ReadRequestHandler.java#L138).

Additionally, `ReadRequestHandler` will now try to get the English version if no content is found. This is so that we can find the English content in the collection when creating a new page in Welsh (that is not already published).

Fixes https://trello.com/c/GpcNQzC3/1163-english-content-overwriting-welsh-content-in-florence

### How to review

- In Florence create content (for example a bulletin) in both English and Welsh
- Approve and publish the collection
- Create a new collection
- Find and edit the previous content in English. Save
- Toggle to Welsh language and find and edit the same content.

Expected: It should load the original Welsh content and not the amended English content in the collection.

### Who can review

Anyone


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/ReadRequestHandler.java
```diff
@@ -173,17 +134,63 @@ public Resource findResource(HttpServletRequest request) throws ZebedeeException
...
+                r -> r.getCollectionContentLength(collectionId, RequestUtils.getSessionId(request), uri),
+                r -> r.getPublishedContentLength(uri));
+    }
+
+    /**
+     * Get a resource in either:
+     * <ol>
+     * <li>the given collection (if any) in the current language</li>
+     * <li>published content in the current language</li>
+     * <li>the given collection (if any) in English</li>
```
cookel2, line 148: I had a play with it and I think the comment in line 148 is not actually true. It seems to be that if there is English content in both master and in collections, and no Welsh content in either, then if someone tries to create a new Welsh page, it will always use the English content from master (the published content) rather than any that is currently in progress. This seems correct but this means that the comment is not quite true.

cookel2 (approved):
It works well so I'm approving it but please see the comment I've added.

zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/ReadRequestHandler.java
rafahop: Thanks for being thorough Ella! 🙂 

It took me a while to figure out why it was behaving as you say when the code seems to be doing what the comment says...

The answer is that the `publishedSupplier` is of type `FileSystemContentReader` whereas the `collectionSupplier` is a `CollectionContentReader`. 
So the `publishedSupplier` will actually try and find the Welsh content and, if it is not there, [will also find the English content](https://github.com/ONSdigital/zebedee/blob/24b122e006a11ebbab38e8377ad9cae95faa628c/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/FileSystemContentReader.java#L399-L401)! Therefore, as far as this handler is concerned, the content was found in published (master) and it never goes into the catch in line 179 which would trigger the bottom 2 scenarios in the list of this comment. It is all really convoluted!!

The scenario in the catch clause is there for when we are creating content for the first time. In that case, we won't have anything in master so neither  `collectionSupplier` nor `publishedSupplier` will find anything, yet we have the English content in the collection (since there is no way of creating Welsh content without the English). Hope this makes sense?

PR #664, opened 2022-11-08
Fix - pick up dataset version URL instead of picking up the dataset URL twice

cookel2:
# What has changed

Previously the doApproval() method, in the ApproveTask class, was passing the following URLs to the 'populate' method of the ReleasePopulator class:
- the dataset version URL as a page of type API_DATASET
- the parent dataset URL as a page of type API_DATASET_LANDING_PAGE
- the parent dataset URL as a page of type API_DATASET_LANDING_PAGE again!

Instead it now passes the following URL to the 'populate' method:
- the dataset version URL as a page of type API_DATASET_LANDING_PAGE

This should resolve the following 2 issues:
1. That the URL being shown on the Dataset Landing Pages was not showing the correct version of the dataset for the relevant release.
2. That the incorrect URL being shown on the Dataset Landing Pages was appearing twice.

# How to review

Ideally create a new release, for either a CMD or cantabular dataset, and check that the dataset URL now takes you to the correct version and only appears once.

The unit tests can be run using this command: make test

The component tests can be run using this command: go test -component

The linter can be run using this command: make lint

# Who can review

Anyone but me


vmilitaru (dismissed):

vmilitaru (approved):
👍

PR #669, opened 2022-11-30
Allow deletion of Welsh content

rafahop:
### What

Fix bug where collections with Welsh content can't be deleted. The Welsh json file remained in the filesystem making the collection impossible to be deleted via Florence.

By using the folder uri instead of the data.json path, the `deletePage` endpoint will now remove all content in that uri, including the `data_cy.json`. Remove all code around deletion of a data json as it is no longer used

https://trello.com/c/tozGt13w/769-deletion-of-welsh-content-within-a-collection-s2

### How to review

Check change makes sense and it won't affect any other functionality.

### Who can review

Anyone but me


cookel2 (changes requested):
Rafa, I tried it locally and it worked correctly in terms of the functionality covered by this PR. However, I found that the develop branch will delete the whole folder structure (if there is a data.json and no data_cy.json to delete). Whereas this branch will only delete the files, as described, and the edition folder that they are contained in.

cookel2 (dismissed):

cookel2 (approved):
It worked successfully for me

PR #675, opened 2022-12-13
Add Dockerfile.local for zebedee-reader

DavidSubiros:
### What

- [Trello card](https://trello.com/c/S2KyzIGX/1024-create-census-hub-stack-using-docker)
- Add `Dockerfile.local` for zebedee-reader

### How to review

- Make sure Dockerfile.local is ok
- You may run the dp-compose stack using [the branch for this PR](https://github.com/ONSdigital/dp-compose/pull/123)

### Who can review

Anyone


zebedee-reader/Dockerfile.local
```diff
@@ -0,0 +1,25 @@
...
+##########################
+
+FROM openjdk:8-jre
+
+# Add the repo source
+WORKDIR /usr/src
+ADD ./target/dependency /usr/src/target/dependency
+ADD ./target/classes /usr/src/target/classes
+
+# Temporary: expose Elasticsearch
```
cookel2, line 13: Why is this marked as temporary?

DavidSubiros: I used the main docker file (cms) as a reference (https://github.com/ONSdigital/zebedee/blob/develop/Dockerfile) - and just replaced the things that needed to be replaced:
- Add reader dependency and classes folders (instead of CMS equivalents)
- Add env vars to configure some parameters

To be honest I don't know if this port needs to be exposed, but I did it so it's consistent with the cms DockerFile.

DavidSubiros: I validated that this port is not being used in any manifest or stack. I will remove it.

redhug1 (approved):
LGTM

PR #685, opened 2023-02-14
Temporarily remove error so that read-only mounted volumne can be tested

cookel2:
### What

I have temporarily removed an error so that I can find out what will happen if it does not get triggered when someone tries to write to the read-only mounted volume in the zebedee container in sandbox. This is required for the following ticket:

https://trello.com/c/MXPuJaOJ/1301-check-data-file-path-is-under-content-and-normalise-it

### How to review

n/a

### Who can review

Anyone but me


cookel2:
Closing as advised by my tech lead

PR #689, opened 2023-03-31
Changed string given to link to allow most recent version of Cantabular dataset to show

JoshSinghONS:
### What

Currently when Zebedee creates a release page it shows a list of Cantabular datasets with links to the first edition of that dataset even if it has been superseded. This code change amends the url which is used to create the link so that it will redirect to the most recent edition of the dataset.

### How to review

LGTM.

### Who can review

Anyone.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/ReleasePopulator.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/ReleasePopulator.java
@@ -121,7 +121,8 @@ public class ReleasePopulator {
     }
 
     private static Link createLink(ContentDetail contentDetail) {
-        Link link = new Link(URI.create(contentDetail.uri));
+        String result = contentDetail.uri.substring(0, contentDetail.uri.indexOf("/versions"));
+        Link link = new Link(URI.create(result));
         link.setTitle(contentDetail.description.title);
         return link;
     }
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/ReleasePopulator.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/ReleasePopulator.java
@@ -89,6 +89,7 @@ public class ReleasePopulator {
     }
 
     private static void addRelatedAPIDataset(Release release, ContentDetail contentDetail) {
+        contentDetail.setUri(contentDetail.uri.substring(0, contentDetail.uri.indexOf("/versions")));
         Link link = createLink(contentDetail);
         if (release.getRelatedAPIDatasets() == null) {
             release.setRelatedAPIDatasets(new ArrayList<>());
@@ -121,8 +122,7 @@ public class ReleasePopulator {
     }
 
     private static Link createLink(ContentDetail contentDetail) {
-        String result = contentDetail.uri.substring(0, contentDetail.uri.indexOf("/versions"));
-        Link link = new Link(URI.create(result));
+        Link link = new Link(URI.create(contentDetail.uri));
         link.setTitle(contentDetail.description.title);
         return link;
     }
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
```diff
@@ -1415,8 +1415,8 @@ public void populateReleaseQuietlyShouldAddLinksToReleasePageForCollectionConten
         collection.complete(publisher1Session, releaseJsonUri, recursive);
         collection.review(publisher2Session, releaseJsonUri, recursive);
 
-        ContentDetail cmdDetail = new ContentDetail("My CMD dataset", "/some/uri", PageType.API_DATASET_LANDING_PAGE);
-        FileUtils.write(collection.getReviewed().getPath().resolve("some/uri/data.json").toFile(),
+        ContentDetail cmdDetail = new ContentDetail("My CMD dataset", "/some/uri/versions/test", PageType.API_DATASET_LANDING_PAGE);
+        FileUtils.write(collection.getReviewed().getPath().resolve("/some/uri/versions/test/data.json").toFile(),
```
jessjenkins, line 1419: Will cause a breaking unit test failure.
```suggestion
        FileUtils.write(collection.getReviewed().getPath().resolve("some/uri/versions/test/data.json").toFile(),
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/ReleasePopulator.java
```diff
@@ -89,6 +89,7 @@ private static void addRelatedDataset(Release release, ContentDetail contentDeta
     }
 
     private static void addRelatedAPIDataset(Release release, ContentDetail contentDetail) {
+        contentDetail.setUri(contentDetail.uri.substring(0, contentDetail.uri.indexOf("/versions")));
```
rafahop, line 92: I would expect the release populator to use the content detail as it comes rather than modify it. As far as I can tell this uri comes from the json file, should the json file not have the `/versions` path in the uri in the first place? This seems more like a hack to me

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
```diff
@@ -1415,8 +1415,8 @@ public void populateReleaseQuietlyShouldAddLinksToReleasePageForCollectionConten
         collection.complete(publisher1Session, releaseJsonUri, recursive);
         collection.review(publisher2Session, releaseJsonUri, recursive);
 
-        ContentDetail cmdDetail = new ContentDetail("My CMD dataset", "/some/uri", PageType.API_DATASET_LANDING_PAGE);
-        FileUtils.write(collection.getReviewed().getPath().resolve("some/uri/data.json").toFile(),
+        ContentDetail cmdDetail = new ContentDetail("My CMD dataset", "some/uri/versions/test", PageType.API_DATASET_LANDING_PAGE);
+        FileUtils.write(collection.getReviewed().getPath().resolve("some/uri/versions/test/data.json").toFile(),
```
rafahop, line 1419: I don't think we need to change the json path here, please revert

```suggestion
        FileUtils.write(collection.getReviewed().getPath().resolve("some/uri/data.json").toFile(),
```

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/tasks/ReleasePopulator.java
rafahop: Also it would be a problem if the url contains `/versions` in any part of the path (and not just a suffix)

JoshSinghONS:
@rafahop yo I've read your comments and for context here is the ticket. 

The more I dug into this the more I felt like this might be a backend issue since I'm messing with some Java files and I'm very unfamiliar with the systems so I concede to the fact your probably right in what your saying but also this is totally out of my wheelhouse and I'm not sure what the path is forward for this ticket then if not what I proposed.

Let me know if you got any ideas from reading the ticket.

PR #694, opened 2023-05-03
Remove uris from kafka service log

lindenmckenzie:
### What

Remove uris from the kafka service log - they sometimes produce log lines that are 450k long. 
I left the log as a whole in because it still seems useful to have it there and without uris should be less problematic. 

### How to review

Check is sensible

### Who can review

Not me.


jondewijones (approved):
LGTM

PR #704, opened 2024-01-10
Feature/update dependencies

MariliaE:
### What

Updated external and internal dependencies to latest minor version.

### How to review
Make sure to use java 1.8 as the current jdk and maven is using it (`mvn --version`)
You have to have experience on how to run zebedee so it can serve content. 
You have to have experience on how to run babage so it can test zebedee reader and zebedee cms. 

run `mvn package`
No errors should appear, the build should run ok and the test cases should run ok

run `mvn versions:display-dependency-updates -DallowMajorUpdates=false`
No libraries should be mentioned here that need updating

create the content folder for zebedee with required content in it
execute ./run-reader.sh - should start zebedee-reader with out errors
execute ./run.sh - should start zebedee cms with out errors
terminate all java running java tasks

Test zebedee servers content to babbage: 

copy content from babage into required location of zebedee, or checkout and build dp-content
run zebedee reader ./run-reader.sh (zebedee)
checkout latest babbage  and execute ./run.sh
checkout compile and run dp-router-api
checkout and compile sixteen to serve stylesheets

go to the content  location : http://localhost:8080/economy/inflationandpriceindices/consumerpriceinflation/2015-04-14  in your browser and you it should display the content without error 500

### Who can review

A member of ONS with skills related to babbage, zebedde, sixteen, dp-content and dp-router-api

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,17 +23,23 @@
         <encoding>UTF-8</encoding>
         <project.build.sourceEncoding>${encoding}</project.build.sourceEncoding>
         <project.reporting.outputEncoding>${encoding}</project.reporting.outputEncoding>
-        <restolino.version>0.6.0</restolino.version>
-        <fasterxml.jackson.version>2.15.0</fasterxml.jackson.version>
+        <restolino.version>0.7.1</restolino.version>
+        <fasterxml.jackson.version>2.16.1</fasterxml.jackson.version>
         <apache.poi.version>3.17</apache.poi.version>
-        <spring.version>5.3.20</spring.version>
-        <dp.logging.version>2.0.0-beta.8</dp.logging.version>
-        <batik.version>1.16</batik.version>
-        <logback.version>1.2.11</logback.version>
+        <spring.version>5.3.31</spring.version>
+        <dp.logging.version>v2.0.0-beta.11</dp.logging.version>
+        <batik.version>1.17</batik.version>
+        <logback.version>1.3.14</logback.version>
     </properties>
 
     <dependencyManagement>
         <dependencies>
+            <!-- Keep logging at the beginning of the file so that org.slf4j dependency is imported at its latest version required by logging to work -->
+            <dependency>
+                <groupId>com.github.onsdigital</groupId>
+                <artifactId>dp-logging</artifactId>
+                <version>${dp.logging.version}</version>
+            </dependency>
 
             <!-- Main -->
             <dependency>
@@ -46,7 +52,7 @@
             <dependency>
                 <groupId>com.github.onsdigital</groupId>
                 <artifactId>dp-cryptolite-java</artifactId>
-                <version>1.5.0</version>
+                <version>1.5.1</version>
             </dependency>
 
             <!-- Any sub modules depending on Zebedee reader and content wrappers should depend on project version as all modules are released together under same version-->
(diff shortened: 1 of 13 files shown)


melsner11Ons (dismissed):
you could mention here in the pom file `<!-- Logging .... --> `that logging needs to be imported first or before restolino or any other package that might has logging in it to use the latest version provided by dp-logging - would help to avoid someone moves it down in the dependency list

melsner11Ons (commented):
Still needs approval from ONS

pom.xml
```diff
@@ -23,17 +23,23 @@
...
-        <logback.version>1.2.11</logback.version>
+        <spring.version>5.3.31</spring.version>
+        <dp.logging.version>v2.0.0-beta.11</dp.logging.version>
+        <batik.version>1.17</batik.version>
+        <logback.version>1.3.14</logback.version>
     </properties>
 
     <dependencyManagement>
         <dependencies>
+            <!-- Logging keep logging at the beginning here as the order of imports does matter to maven-->
```
lindenmckenzie, line 37:
```suggestion
            <!-- Logging - keep logging at the beginning here as the order of imports does matter to maven-->
```

lindenmckenzie: Is there something in particular it's clashing with if you put it lower down?

lindenmckenzie (commented):
Some very minor comments - I'll test this locally next.

dionisislolas (commented):
When I run mvn versions:display-dependency-updates -DallowMajorUpdates=false
The following dependencies in Dependency Management have newer versions:
ch.qos.logback:logback-classic ...................... 1.3.14 -> 1.4.14

Is this expected ?

MariliaE:
> When I run mvn versions:display-dependency-updates -DallowMajorUpdates=false The following dependencies in Dependency Management have newer versions: ch.qos.logback:logback-classic ...................... 1.3.14 -> 1.4.14
> 
> Is this expected ?

@dionisislolas this is expected yes, and the reason has been documented at the end of this [confluence page ](https://confluence.ons.gov.uk/display/DIS/Java+repositories). 

ch.qos.logback:logback-classic supports JDK 8 only on logback version 1.3.x ([logback version strategy](https://logback.qos.ch/dependencies.html)), so it's been updated to the latest 1.3.x version, which is [1.3.14](https://mvnrepository.com/artifact/ch.qos.logback/logback-classic)

pom.xml
MariliaE: We found that dp-logging had a library missing _org.slf4j_ which required all java applications to add this dependency so that they would work. What we did is we added this dependency in dp-logging to be pulled in from there and avoid having to include it in every application. However, restolino is also using _org.slf4j_ but an older version of slf4j-api since we are only doing minor updates. By changing the order of dependencies and putting dp-logging above restolino the build then worked as it's using version 2.0.10 which is being pulled from dp-logging.

lindenmckenzie: Sure - so can you note this in the comment so we know exactly why it's being kept there and what we should avoid putting it below in future.

lindenmckenzie:
@MariliaE would be sensible to comment about the logback version in the pom.xml file as well.

lindenmckenzie (approved):

run.sh
```diff
@@ -32,8 +32,9 @@ java $JAVA_OPTS \
  -Drestolino.classes=$RESTOLINO_CLASSES \
  -Drestolino.packageprefix=$PACKAGE_PREFIX \
  -DSTART_EMBEDDED_SERVER=N \
-  -javaagent:zebedee-cms/target/dependency/aws-opentelemetry-agent-1.32.0.jar \
-  -Dotel.propagators=tracecontext,baggage \
+ -javaagent:zebedee-cms/target/dependency/aws-opentelemetry-agent-1.31.0.jar \
```
lindenmckenzie, line 35: Why is this downgrading the otel agent? I thought you put in the upgrade to sort some issues?

lindenmckenzie (commented):
Clarification required on this.

run.sh
MariliaE: Thanks for catching this, I made the update now and tested Zebedee with Babbage locally which shows the content being displayed without any errors!

lindenmckenzie (approved):

PR #708, opened 2024-01-31
Release/2.20.1

MariliaE:
### What
Release 2.20.1 includes:

- Update internal and external dependencies to latest minor version
- Disabling otel

### How to review

Check release changes are okay.

### Who can review

Anyone

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,17 +23,23 @@
         <encoding>UTF-8</encoding>
         <project.build.sourceEncoding>${encoding}</project.build.sourceEncoding>
         <project.reporting.outputEncoding>${encoding}</project.reporting.outputEncoding>
-        <restolino.version>0.6.0</restolino.version>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -217,7 +216,7 @@
             <dependency>
                 <groupId>org.elasticsearch</groupId>
                 <artifactId>elasticsearch</artifactId>
-                <version>2.1.1</version>
+                <version>2.4.6</version>
```
gedge, line 219: there seems to be a problem in Sandbox where the reindex is not working

this ES lib upgrade may cause the issue (a `jackson-dataformat-yaml` depedency clash causing a problem where `snakeyaml` 1.15 is now being used (and failing at reindex time) where we used to have `snakeyaml` 2.2 working

I am attempting a quick fix, but this release should not go in until it works with reindex

PR #714, opened 2024-03-12
Send request to Legacy Cache API on approval

jmgq:
### What

We are currently sending requests to the Legacy Cache API when we receive a `PUBLISHED` event, but we should be sending them when we receive an `APPROVED` event. This PR fixes this. The reason why it was done this way is that we didn't think that we could get the publishing time on `APPROVED` events, we thought that the time we get with those events was the time at which the changes were approved. However, we have now verified that the date returned by `collection.getDescription().getPublishDate()` when receiving an `APPROVED` event is the correct publishing date, and not the approval date/time, like we originally thought.

Therefore, we're now sending a PUT request to the Legacy Cache API when we receive an `APPROVED` event, and we're ignoring the `PUBLISHED` events, as they're now redundant, and they are sent at the publishing time, which is too late for the Legacy Cache Proxy to be able to set the calculated `max-age` directive.

### How to review

Run the ONS stack locally, ensure that the Legacy Cache API is running, and set the `ENABLE_LEGACY_CACHE_API` feature flag. Then use Florence to create or update collections. When you approve the changes, you should see how the API immediately receives a call from Zebedee.

### Who can review

Anyone at ONS.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PublishNotification.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PublishNotification.java
@@ -38,13 +38,12 @@ public class PublishNotification {
     }
 
     public PublishNotification(Collection collection, List<String> urisToUpdate, List<ContentDetail> urisToDelete) {
-        // Delay the clearing of the cache after publish to minimise load on the server while publishing.
-        Date clearCacheDate = new DateTime(collection.getDescription().getPublishDate())
-                .plusSeconds(Configuration.getSecondsToCacheAfterScheduledPublish()).toDate();
-
         if (Configuration.isLegacyCacheAPIEnabled()) {
-            this.legacyCacheApiPayloads = new LegacyCacheApiPayloadBuilder.Builder().collection(collection).publishDate(clearCacheDate).build().getPayloads();
+            this.legacyCacheApiPayloads = new LegacyCacheApiPayloadBuilder.Builder().collection(collection).build().getPayloads();
         } else {
+            // Delay the clearing of the cache after publish to minimise load on the server while publishing.
+            Date clearCacheDate = new DateTime(collection.getDescription().getPublishDate())
+                    .plusSeconds(Configuration.getSecondsToCacheAfterScheduledPublish()).toDate();
             this.payload = new NotificationPayload(collection.getDescription().getId(), urisToUpdate, urisToDelete, clearCacheDate);
         }
     }
@@ -110,7 +109,7 @@ public class PublishNotification {
     }
 
     private void sendRequestToLegacyCacheApi(EventType eventType) {
-        if (eventType.equals(EventType.PUBLISHED) || eventType.equals(EventType.UNLOCKED)) {
+        if (eventType.equals(EventType.APPROVED) || eventType.equals(EventType.UNLOCKED)) {
             info().data("eventType", eventType.name()).log("sending request to Legacy Cache API");
 
             removePublishDateForUnlockedEvents(eventType);
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/legacycacheapi/LegacyCacheApiPayloadBuilder.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/legacycacheapi/LegacyCacheApiPayloadBuilder.java
@@ -21,7 +21,6 @@ public class LegacyCacheApiPayloadBuilder {
 
     public static class Builder {
         private Collection collection;
-        private Date publishDate;
         private final Map<String, LegacyCacheApiPayload> payloadsByPath = new HashMap<>();
 
         public Builder() {}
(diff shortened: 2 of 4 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PublishNotification.java
```diff
@@ -38,13 +38,12 @@ public class PublishNotification {
     }
 
     public PublishNotification(Collection collection, List<String> urisToUpdate, List<ContentDetail> urisToDelete) {
-        // Delay the clearing of the cache after publish to minimise load on the server while publishing.
-        Date clearCacheDate = new DateTime(collection.getDescription().getPublishDate())
-                .plusSeconds(Configuration.getSecondsToCacheAfterScheduledPublish()).toDate();
-
         if (Configuration.isLegacyCacheAPIEnabled()) {
-            this.legacyCacheApiPayloads = new LegacyCacheApiPayloadBuilder.Builder().collection(collection).publishDate(clearCacheDate).build().getPayloads();
+            this.legacyCacheApiPayloads = new LegacyCacheApiPayloadBuilder.Builder().collection(collection).build().getPayloads();
```
melsner11Ons, line 42: Should we log publishdate here ?

melsner11Ons (commented):
Looks good to me

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PublishNotification.java
jmgq: We're already logging each individual payload when we send them to the API, and that includes the publishing date, so I think that covers it.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PublishNotification.java
```diff
@@ -110,7 +109,7 @@ private void sendNotificationToWebsite(EventType eventType) {
     }
 
     private void sendRequestToLegacyCacheApi(EventType eventType) {
-        if (eventType.equals(EventType.PUBLISHED) || eventType.equals(EventType.UNLOCKED)) {
+        if (eventType.equals(EventType.APPROVED) || eventType.equals(EventType.UNLOCKED)) {
```
gedge, line 112: I confess to not knowing the answer to this, but (tldr: _why not both?_)

I understand that the `PUBLISHED` event causes the upcoming publish time to be removed from the cache, thus tidying the cache of past publish-dates.

So, 

- yes, you want `APPROVED` to add the publish-date to the cache
- but I think you also want `PUBLISHED` to tidy-up (much like `UNLOCKED`)

unless you're tidying up in a way I don't understand

lindenmckenzie: I would concur with @gedge above - this would reduce calculation time in the proxy if it just isn't there. I would also agree that I believe this is what happens at present.

jmgq: There is no concept of tidying up the cache of past publish-dates in the Legacy Cache API. These three scenarios have the same result (a default cache time):

- The API does not have a resource for a certain path (i.e. the API returns a 404).
- The resource does not have a publish date.
- The resource has a publish date, but it's in the past (and the publish expiry offset flag is off).

The calculation time in the proxy for all those scenarios is pretty much the same. Well, strictly speaking, there are fewer calculations to process a 404 than a past publish date, but the three scenarios require you to make a network call to the API, which is where the bulk of the time will go.

If we start also processing the `PUBLISHED` events, what will happen is that we will start making unnecessary requests to the API. Unless I'm mistaken, for a `PUBLISHED` event to happen it must have been approved first, so we would have already received an `APPROVED` event beforehand. By processing `PUBLISHED` events we would be repeating ourselves. Furthermore, the `PUBLISHED` events seem to be received at publishing time, which is already too late for the Proxy to edit the `max-age`.

Happy to be corrected on any of this, though.

gedge: my understanding is this

## Scenario 1: Currently in Prod (before legacy-cache)

1. collection is `APPROVED`
   - babbage is informed and stores the PublishTime in a cache
2. collection is later `PUBLISHED` or `UNLOCKED`
   - babbage is informed and removes the PublishTime from a cache

so, the cache (elasticsearch?) is cleared whether or not the collection is Published

## Scenario 2: using the legacy-cache

1. collection is `APPROVED`
   - legacy-cache-API is informed and stores the PublishTime in the cache
2. collection is later `UNLOCKED`
   - legacy-cache-API is informed and removes the PublishTime from the cache (or not)

In this scenario, the 'cache' is mongodb/docdb, I believe, and _does not get cleared of successful PublishTimes_ (there's no tidy-up for `PUBLISHED`)

Now, this may be because _max-age_ remains low for a limited time after `PUBLISHED` (short-term post-publication), but it seems the cache will accumulate indefinitely (untidily) - and I want this clarified

Hope that makes sense.

jmgq: Thank you, it makes sense.

> 2. collection is later `UNLOCKED`
> legacy-cache-API is informed and removes the PublishTime from the cache (or not)

The publish time is always removed from the cache when an `UNLOCKED` event is received. We have only seen `UNLOCKED` events when we manually click the "Unlock" button. Also, to clarify, when I say that the publish time is always removed, I mean that the `release_time` attribute of the Cache Time resource in the API is set to `nil`. Resources are never deleted.

> In this scenario, the 'cache' is mongodb/docdb, I believe, and does not get cleared of successful PublishTimes (there's no tidy-up for PUBLISHED)

That's right.

> Now, this may be because max-age remains low for a limited time after PUBLISHED (short-term post-publication)

We have a feature flag around this. If that feature flag is on, then you are correct: for the next 3 minutes after a release, the cache time returned by the proxy is the "short cache time" (10s by default). If the feature flag is off (which it is, by default), then this functionality does not apply, and instead we return a default cache time (900s) after release.

> but it seems the cache will accumulate indefinitely (untidily)

I'm not sure what this means. Do you mean that since we're not setting the publish time to `nil` after publishing, that the API will return a publish time that's in the past for a certain resource (rather than returning no publish time at all)? If that's what you mean, then yes, that's how it works. This is not a problem because both scenarios (a resource with no release time, and a resource with release time in the past) produce the same output (a default cache time).

gedge: I think we're getting there, thanks. 😄

To address your last paragraph: I'm not at all concerned about the cache time being wrong. Functionally, it appears to be right to the outside user.

But, why should docdb accumulate data that is no longer useful? Why not clear docdb of each published collection/time:
 
- after its publication (if the mentioned _post-publish window_ feature flag is *off*)?
- or after the publish time plus the "short cache time" (if the flag is *on*)?

Why set the time to `nil`?

To restate: I'm more concerned about docdb accumulating useless data, over time, and am trying to find out why we aren't clearing it up.

jmgq: Yes, that makes sense. I'm not sure why it was designed like this in the first place. James would be the right person to ask. Perhaps it was done this way to leave a trace, a history.

But if you'd rather have all the published resources removed, then that could be a future feature. But at the moment, the API does not even accept requests with the `DELETE` method.

gedge (approved):
on the basis that the spec says this is right

PR #717, opened 2024-03-25
Enable debugging in docker

dionisislolas:
### What

Enable debugging in docker

### How to review

Run auth stack and check if zebedee can be debugged.

### Who can review

Anyone

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ADD ./zebedee-cms/target/classes /usr/src/target/classes
 EXPOSE 9200
 
 # Update the entry point script
-ENTRYPOINT java -Xmx2048m \
+ENTRYPOINT java -Xmx2048m -agentlib:jdwp=transport=dt_socket,address=8002,server=y,suspend=n \
           -Drestolino.classes=target/classes \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -9,7 +9,7 @@ ADD ./zebedee-cms/target/classes /usr/src/target/classes
 EXPOSE 9200
 
 # Update the entry point script
-ENTRYPOINT java -Xmx2048m \
+ENTRYPOINT java -Xmx1204m -Xdebug -Xrunjdwp:transport=dt_socket,address=8002,server=y,suspend=n \
```
lindenmckenzie, line 12: Few comments - 
1. why the reduction in heap space? 
2. seems to be some pointers that -Xrunjdwp is for pre Java 5. `-agentlib:jdwp` is the more modern way.
3. you may want to make the address configurable with a default for future conflicts
4. please make sure this is documented in the readme

lindenmckenzie (commented):
Some comments.

Dockerfile
dionisislolas: Made some changes and seems to be working now also with the port configurable.

BarathaAberathne (approved):
Branch tested using IntelliJ and debugger works Successfully!

PR #726, opened 2024-04-25
Feature/dependency bump and vulnerability fix

Oluwawunmi:
### What
Removed some vulnerabilities and bumped some dependencies

How to review
Pull and check updates.

### Who can review
Anyone

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -436,6 +436,11 @@
                             <artifactId>lucene-queryparser</artifactId>
                             <version>5.5.4</version>
                         </exlude>
+                        <exlude>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -436,6 +436,11 @@
                             <artifactId>lucene-queryparser</artifactId>
                             <version>5.5.4</version>
                         </exlude>
+                        <exlude>
```
jessjenkins, line 439: There's no Trello link fo this new audit exclusion and it appears to be related to the one above which is actually unrelated.

jessjenkins (approved):

PR #744, opened 2024-07-23
Use distinct_hosts

redhug1:
### What

Reverting to this instead of spread by node.unique.id as zebedee seen twice on one web_mount in staging.

### How to review

Visual inspection

### Who can review

Anyone


lindenmckenzie (approved):
Reverts to pre-spread state.

PR #760, opened 2024-09-13
Created function to find the dataset version

dionisislolas:
### What

Created function to find the dataset version

### How to review

- Run the stack dp-compose/v2/stacks/static-files-with-auth
- In florence create a manual collection to change/update a dataset.
- Change the content of the file to make sure the correct file is being uploaded.
- Check correct files are uploaded in the private bucket on dataset approval.
- Check correct files are in the public bucket on dataset publication.

Please note:

- Check both timeseries and non-timeseries datasets.

### Who can review

Anyone


BarathaAberathne (approved):
Approving without testing in localhost. since the changes are protected using a feature flag this changes won't effect other functionalities.

PR #763, opened 2024-09-23
make sure we pick the correct dataset version

BarathaAberathne:
### What

Addressed issues from the previous PR
making sure we are picking the correct dataset version

### How to review

Check if changes makes sense

### Who can review

@anyone

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
@@ -499,7 +499,7 @@ public class ApproveTask implements Callable<Boolean> {
         return finalPath;
     }
 
-    protected String pickCorrectDatasetVersion(String currentDatasetVersion, String defaultDatasetVersion) throws IllegalArgumentException {
+    protected String pickCorrectDatasetVersion(String currentDatasetVersion, String defaultDatasetVersion) {
         if (defaultDatasetVersion == null || currentDatasetVersion == null) {
             throw new IllegalArgumentException("input can't be null");
         }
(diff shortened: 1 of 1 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java
```diff
@@ -526,16 +526,17 @@ protected String incrementDatasetVersionByOne(String datasetVersion) {
         return letter + Integer.toString(number);
     }
 
-    protected String findCorrectDatasetVersion(List<String> listOfUris) {
+    protected String findCorrectDatasetVersion(List<String> listOfUris) throws IllegalArgumentException {
```
lindenmckenzie, line 529: so this exception isn't getting caught - so won't this just cause the application to fall over? I may be misremembering my java.

dionisislolas: Sorry Linden we made a mistake. The IllegalArgumentException that this method throws is an Unchecked Exception so we don't have to declare it. So we removed the throws clause in the method signature. And also the exception that this method throws will be caught further up in the function hierarchy, namely on this catch clause
[https://github.com/ONSdigital/zebedee/blob/387c589255f632e9c3a0a3330b0ae8c69302e50[…]a/com/github/onsdigital/zebedee/model/approval/ApproveTask.java](https://github.com/ONSdigital/zebedee/blob/387c589255f632e9c3a0a3330b0ae8c69302e505/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/approval/ApproveTask.java#L184)

lindenmckenzie: Great, thanks Dionisi

lindenmckenzie (approved):

PR #798, opened 2025-06-11
Stop latest being set if migrationLink is present

lindenmckenzie:
### What

Stop `latestRelease` being set if `migrationLink` is present on latest item in a series.

### How to review

Check looks ok. 

To test you can:

- find your zebedee content store
- alter the latest edition of an editorial series to have a `migrationLink` in the description block
- query for the uri of the edition on the /data endpoint (/data?uri=${uri}
- see that `latestRelease` is no longer set

Regression test against those with it not set. 

### Who can review

Not me.

### DIFF
--- a/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/FileSystemContentReader.java
+++ b/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/FileSystemContentReader.java
@@ -196,7 +196,9 @@ public class FileSystemContentReader implements ContentReader {
             Path parent = contentPath.getParent();
             assertIsEditionsFolder(parent);
             page = resolveLatest(contentPath);
-            page.getDescription().setLatestRelease(true);
+            if (page.getDescription().getMigrationLink() == null || page.getDescription().getMigrationLink().isEmpty()){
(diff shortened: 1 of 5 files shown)


zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/FileSystemContentReader.java
```diff
@@ -196,7 +196,9 @@ public Page getLatestContent(String path) throws ZebedeeException, IOException {
             Path parent = contentPath.getParent();
             assertIsEditionsFolder(parent);
             page = resolveLatest(contentPath);
-            page.getDescription().setLatestRelease(true);
+            if (page.getDescription().getMigrationLink() == null || page.getDescription().getMigrationLink().isEmpty()){
```
Oluwawunmi, line 199: Not a big deal but we can use apache-commons string utils here and just check if blank.

lindenmckenzie: resolved.

Oluwawunmi (approved):
LGTM

PR #806, opened 2025-07-29
Integrate search deletion with Zebedee

Oluwawunmi:
### What

When a user deletes something in zebedee - when published, issue a `search-content-deleted` Kafka message during the POST-PUBLISH stage.

### How to review

Sense check
Ensure tests Pass

### Who can review

!me


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/PostPublisher.java
```diff
@@ -392,4 +413,27 @@ private static void sendMessage(Collection collection, List<String> uris, String
             notifier.sendCollectionAlarm(collection, channel, "Failed to send content-updated kafka events", e);
         }
     }
+
+    private static void sendContentDeletedEventsToKafka(Collection collection, List<String> uris) {
+        String traceId = defaultIfBlank(MDC.get(TRACE_ID_HEADER), UUID.randomUUID().toString());
```
jessjenkins, line 418: I know it's copied and pasted from a bad precedent so not blaming you, but I do not like this at all. It's not the PostPublisher's job to be generating traceIds. Not only that, but as we are reating a new one, it could mean that different parts of the request have different IDs. (eg. the content-updated ones will have a different ID from the content-deleted ones). If for some reason this request doesn't have a traceID, it's OK to not send one on. It should be generated on message consumption in the downstream app if it's missing. In any case the requests always have a traceID anyway as it's set up by the [MDCFilter](https://github.com/ONSdigital/zebedee/blob/61741a5329564a1ea8e74c2c027b2a0e1c40bd12/zebedee-reader/src/main/java/com/github/onsdigital/zebedee/reader/api/filter/MDCFilter.java#L28)  on request handling.

```suggestion
        String traceId = MDC.get(TRACE_ID_HEADER);
```
Even better, abstract the MDC stuff completely into some other class entirely so anywhere in th code can deal with MDC internals and headers etc in a more managed way…
```suggestion
        String traceId = SomeRequestContextClass.getTraceId();
```

jessjenkins (approved):

PR #807, opened 2025-08-02
process the list of redirects on collection post-publish

Oluwawunmi:
### What

Implemented publishRedirects in RedirectServiceImpl to handle CREATE, UPDATE, and DELETE redirect actions after a collection reaches POST-PUBLISH.

### How to review

Sense check: Review the implementation of publishRedirects in RedirectServiceImpl.
Confirm all tests.

### Who can review

!me


lindenmckenzie (commented):
Minor piece of consistency - happy to flex either way, but we should be consistent between us.

PR #841, opened 2026-03-16
Feature/add automated collections

lindenmckenzie:
### What

This PR makes several changes to the zebedee collections and permissions approaches. 

Functionality added:

- new collection type: "automated"
- adding checking for permission by collection type. This goes into every part of zebedee virtually.
- adding a new permission level of 'legacy:self-approve'
- allowing self approval if the collection is of type automated
- various linting / removal of old unused things

The intent of this is to make it possible for a user with the 'legacy:self-approve' permission able to approve collections with the type of 'automated'. It is also intended to be used to limit the scope of other 'legacy' permissions by collection type.

### How to review

The easiest way to test this is to run with the legacy-core-publishing stack in dp-compose. 

You will also need to be using: https://github.com/ONSdigital/dp-permissions-api/pull/174 for the population of the permissions api. 

When everything is running you can use the dis-migration-service service auth token to:

- create a collection (POST /collection) with type 'automated'
- add content to that collection (POST `/content/{collectionID}?uri={aValidURI}/data.json`)
- complete the content in that collection (POST `/complete/{collectionID}?uri={aValidURI}/data.json`)
- approve the content in that collection (POST `/review/{collectionID}?uri={aValidURI}/data.json`)
- approve the collection (POST `/approve/{collectionID}`)
- publish the collection (POST `/publish/{collectionID}`)

The above steps are intended to be done in sequence - out of sequence you can:
- unlock the collection (POST `/unlock/{collectionID}`) - this is only possible if the collection is in the queue (i.e. all content is approved and the collection is approved
- delete the collection (DELETE `/collection/{collectionID}`) - this is only possible if the collection is empty and not in the queue

For regression testing the same steps should be done with a user, including attempting to approve their own collection edits. 

N.B. without Florence changes, the automated collections will appear in the UI and be editable by users. 

### Who can review

Not me.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Transfer.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Transfer.java
@@ -17,35 +17,44 @@ import java.nio.file.Path;
 
 /**
  * Created by kanemorgan on 24/03/2015.
+ * TODO:I don't believe this is used and so could be removed LM 16/03/2026
  */
 @Api
 public class Transfer {
 
     /**
-     * Moves files between collections using the endpoint <code>/Transfer/[CollectionName]</code>
+     * Moves files between collections using the endpoint
+     * <code>/Transfer/[CollectionName]</code>
      *
-     * @param request This should contain a X-Florence-Token header for the current session
-     * @param response <ul>
-     *                 <li>If the either collection does not exist:  {@link HttpStatus#NOT_FOUND_404}</li>
-     *                 <li>If the file does not exist:  {@link HttpStatus#NOT_FOUND_404}</li>
-     *                 <li>If user not authorised to transfer:  {@link HttpStatus#UNAUTHORIZED_401}</li>
-     *                 <li>A file with that name already exists in the second collection:  {@link HttpStatus#CONFLICT_409}</li>
-     * @param params A {@link TransferRequest} object
+     * @param request  This should contain a X-Florence-Token header for the current
+     *                 session
+     * @param response
+     *                 <ul>
+     *                 <li>If the either collection does not exist:
+     *                 {@link HttpStatus#NOT_FOUND_404}</li>
+     *                 <li>If the file does not exist:
+     *                 {@link HttpStatus#NOT_FOUND_404}</li>
+     *                 <li>If user not authorised to transfer:
+     *                 {@link HttpStatus#UNAUTHORIZED_401}</li>
+     *                 <li>A file with that name already exists in the second
+     *                 collection: {@link HttpStatus#CONFLICT_409}</li>
+     * @param params   A {@link TransferRequest} object
      * @return success true/false
      * @throws IOException
      */
     @POST
-    public boolean move(HttpServletRequest request, HttpServletResponse response, TransferRequest params) throws IOException {
+    public boolean move(HttpServletRequest request, HttpServletResponse response, TransferRequest params)
+            throws IOException {
         // user has permission
         Session session = Root.zebedee.getSessions().get();
-        if (!Root.zebedee.getPermissionsService().canEdit(session)){
-            response.setStatus(HttpStatus.UNAUTHORIZED_401);
+        if (!Root.zebedee.getPermissionsService().canEdit(session)) {
+            response.setStatus(HttpStatus.FORBIDDEN_403);
             return false;
         }
 
         // get the source collection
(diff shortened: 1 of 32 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Transfer.java
```diff
@@ -17,35 +17,44 @@
 
 /**
  * Created by kanemorgan on 24/03/2015.
+ * TODO:I don't believe this is used and so could be removed LM 16/03/2026
```
jessjenkins, line 20: Note, there appears to be a function defined in florence for it by I can't see it being used. 
https://github.com/ONSdigital/florence/blob/21307a1320739c541acf28278854427905a56619/src/legacy/js/functions/_transfer.js#L1

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Version.java
```diff
@@ -88,6 +98,15 @@ public boolean delete(HttpServletRequest request, HttpServletResponse response)
         Collection collection = Collections.getCollection(request);
         String uri = request.getParameter("uri");
 
+        CollectionDescription collectionDescription = collection.getDescription();
+        if (collectionDescription == null) {
+            throw new InternalServerError("Collection description is missing");
+        }
+
+        if (!Root.zebedee.getPermissionsService().canEdit(session, collectionDescription.getType())) {
+            throw new ForbiddenException("You are not authorised to edit content for this collection type.");
```
jessjenkins: "Delete" maybe?
```suggestion
            throw new ForbiddenException("You are not authorised to delete content for this collection type.");
```

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
```diff
@@ -890,7 +897,59 @@ public void shouldNotReviewAsPublisher() throws IOException, ZebedeeException {
...
+        assertTrue(reviewed);
+        Path reviewedPath = builder.collections.get(1).resolve(Collection.REVIEWED);
+        assertTrue(Files.exists(reviewedPath.resolve(uri.substring(1))));
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void shouldNotSelfApproveForNonAutomatedCollection() throws Exception {
+
+        // Given a non-automated collection
+        collection.getDescription().setType(CollectionType.manual);
```
jessjenkins: Just wondering if it's worth covering scheduled and manual collections separately rather than together as 'non-automated'. I know the way the code is written this is covered the same way, but the tests would ensure this even if someone chnges it in the future?

lindenmckenzie: Resolved.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Version.java
lindenmckenzie: Resolved.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collections.java
```diff
@@ -554,11 +568,16 @@ public void delete(Collection collection, Session session)
...
+        }
+
+        CollectionDescription collectionDescription = collection.getDescription();
+        if (collectionDescription == null) {
+            throw new BadRequestException("Collection description not found for collection");
         }
 
         // User has permission
-        if (!permissionsService.canEdit(session)) {
+        if (!permissionsService.canEdit(session, collectionDescription.getType())) {
```
jessjenkins, line 580: Found an interesting bug here while testing. Although this code is correctly throwing appropriate exceptions which should give the expected status codes, [the exception is being caught in the api class and re-throwing as an InternalServerError exception](https://github.com/ONSdigital/zebedee/blob/b664dea68be583ddc64985e69415961d76f8b247/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java#L330) so is always returning 500.

lindenmckenzie: As agreed - this is existing behaviour for a bad request and so not addressing here.

jessjenkins (approved):

PR #844, opened 2026-04-14
Feature/pass file uris

lindenmckenzie:
### What

Added functionality to pass file uris to the cache api.

I've moved some tests around as they were misleading what they were testing. 

### How to review

Check looks ok - to test it you can run with:
- https://github.com/ONSdigital/dp-compose/pull/295
- https://github.com/ONSdigital/dp-legacy-cache-api/pull/53

If you schedule a collection with a file in and approve it you should now see the file in the cachetimes database.

### Who can review

Not me.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/legacycacheapi/LegacyCacheApiPayloadBuilder.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/legacycacheapi/LegacyCacheApiPayloadBuilder.java
@@ -1,6 +1,8 @@
 package com.github.onsdigital.zebedee.model.publishing.legacycacheapi;
 
 import com.github.onsdigital.zebedee.model.Collection;
+import com.github.onsdigital.zebedee.util.URIUtils;
+
 import org.apache.commons.lang.StringUtils;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/legacycacheapi/LegacyCacheApiPayloadBuilder.java
```diff
@@ -75,8 +83,35 @@ private PayloadType getPayloadType(String uriToUpdate) {
             return PayloadType.NONE;
         }
 
+        /**
+         * Adds a payload for a specific file if the URI matches a file type.
```
cookel2, line 87: Is this comment correct?  I cannot see where the code is matching the URI against a file type.

cookel2 (commented):
Just a comment on the syntax. I still need to test it though.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/legacycacheapi/LegacyCacheApiPayloadBuilder.java
lindenmckenzie: The relevant line is:

```
URIUtils.isLastSegmentFileExtension(uriToUpdate)
```

cookel2: Ok thanks. I did wonder if it was that line but was unsure. I guess that the uriToUpdate must be indicating a certain file type then? I had incorrectly thought that the  "isLastSegmentFileExtension" method was just looking to see if the last segment contained a file extension or not.

cookel2 (dismissed):
LGTM and it works nicely, thanks

cookel2 (approved):
I retested it and it still worked successfully

PR #850, opened 2026-04-23
Collection locking

lindenmckenzie:
### What

This is a PR for introducing some level of "collection locking" to zebedee. 

This was brought about by another service acting in parallel and making very quick requests, causing the collection description JSON to be overridden and therefore data loss was occurring. This is because a write lock is only acquired during the collection.save() function which happens after all the other processing that goes on, that can be slow due to the interaction with disk. 

Initially I approached this just as a collection description json problem and dealt with it via storing an eTag for comparison in memory - in practice, this leads to concurrent requests to save content being actioned (i.e. updating the content files) but then the collection description json update being rejected with a 409 which can't then be retried as the update to the files has already been made. 

Therefore I have moved to acquiring a writelock at the start of any collection writing activity, which is then released at the end of the call. This is a deliberately non-destructive change (i.e. the `writeable` property does not block writing) for backwards compatibility.

### How to review

Good question - you can test with the migration stack and run the test job. With this change, all items posted should now make it into the collection json. Before it was around 3-4 out of six. This is more apparent if the migration service completes and reviews the items too. To do completion / approval you can use this branch: https://github.com/ONSdigital/dis-migration-service/pull/92

You can also work with it in a more standard mode with legacy-core-publishing and try lots of repeated requests. 

### Who can review

Not me.


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -215,50 +215,59 @@ public CollectionDescription update(HttpServletRequest request, HttpServletRespo
...
+                    .fromTo(collection.getPath().toString(), updatedCollection.getPath().toString())
+                    .actionedBy(session.getEmail())
+                    .log();
 
-        info().user(session.getEmail())
-                .collectionID(collectionId)
-                .log("update collection endpoint: request completed successfully");
-        return updatedCollection.getDescription();
+            // close collection writelock
+            collection.close();
```
jessjenkins: Could double unlocking cause a conflict? Consider this scenario…
- Thread 1 locks the collection
- Thread 1 unlocks the collection here
- Thread 2 locks the collection very quickly while thread 1 is still returning the collection
- Thread 1 unlocks the re-locked collection via its close() in the finally block
- Thread 3 locks the collection while thread 2 is still working
- Threads 2 and 3 get into classic, old-school zebedee conflict funkiness as before

jessjenkins: NB. might be worth some sort of local `closed` boolean or somehting to prevent double unlocking

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/api/Collection.java
```diff
@@ -286,38 +295,46 @@ public boolean deleteCollection(HttpServletRequest request, HttpServletResponse
...
+            deleteCollection(collection, session);
 
-        Audit.Event.COLLECTION_DELETED
-                .parameters()
-                .host(request)
-                .collection(collection)
-                .actionedBy(session.getEmail())
-                .log();
+            // close collection writelock
+            collection.close();
```
jessjenkins: Double unlock risk (see other comment)

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ApproveTest.java
```diff
@@ -1,36 +1,85 @@
...
+    @Mock
+    private com.github.onsdigital.zebedee.model.Collections collections;
+
+    @Mock
+    private Collection collection;
+
+    @Mock
+    private Session session;
+
+    private Approve api;
```
jessjenkins: Is `api` a little generic here, could we use `apiApprove` or something like that to show it's the instance of the api.Approve under test.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/CollectionTest.java
```diff
@@ -67,6 +73,9 @@ public class CollectionTest extends ZebedeeAPIBaseTestCase {
     @Mock
     private User user;
 
+        @Mock
+        private Zebedee zebedee;
+
```
jessjenkins: Indentation too wide?

jessjenkins: Actually there are further indentation weirdness in this file so maybe reformat it all.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ApproveTest.java
jessjenkins: Similar issue in other tests all using `api` as the instance under test.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionsTest.java
```diff
@@ -50,7 +50,11 @@
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
```
jessjenkins: Unused imports?

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/model/CollectionTest.java
```diff
@@ -550,6 +553,94 @@ public void shouldCreate() throws IOException {
...
+        Path collectionPath = builder.collections.get(1);
+        Path collectionJsonPath = collectionPath.getParent().resolve(collectionPath.getFileName() + ".json");
+
+        FileUtils.write(collectionJsonPath.toFile(), "not valid json", Charset.defaultCharset());
+
+        try {
+            new Collection(collectionPath, zebedee, true);
+        } catch (IOException expected) {
+            // expected
+        }
```
jessjenkins: Rather than just assuming it will throw, why not assert it?

```suggestion
        assertThrows(IOException.class,() -> {
            new Collection(collectionPath, zebedee, true);
        });
```

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/ApproveTest.java
lindenmckenzie: yep done.

zebedee-cms/src/test/java/com/github/onsdigital/zebedee/api/CollectionTest.java
lindenmckenzie: done.

jessjenkins: Not done. I think you reformatted model.Collection. This is api.Collection ;-)

jessjenkins (approved):

PR #854, opened 2026-05-18
Add guard clause for null versions

lindenmckenzie:
### What

Zebedee has an 'experimental' missing versions checker during the approval step on a collection. 

Unfortunately some datasets don't have the versions set at all - this guards against the versions coming back as null.

### How to review

You can test this with the migration stack.

You will need to:

- pull this branch
- make build in this repo
- set ENABLE_DATASET_VERSION_VERIFICATION in zebedee's manifest
- SERVICE=zebedee make refresh
- alter the test dataset so it does not have a `versions` attribute in the latest `dataset` page for the series
- run the migration
- approve the migration

it now moves to `published`

### Who can review

Not me.


janderson2 (approved):
LGTM 👍

PR #856, opened 2026-05-21
filter manifest from publish

aqdas0307:
### What

[filter manifest from publish](https://officefornationalstatistics.atlassian.net/browse/DIS-4893?atlOrigin=eyJpIjoiYzNhOGY1MDI5ZTI5NDU5ZTlmNDE2NTliYzgzMzM3NTYiLCJwIjoiaiJ9)
-> as opposed to filtering the versions out (which is happening already within manifest)

### How to review

sense check
ensure tests pass

### Who can review

not me

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
@@ -124,7 +124,6 @@ public class Publisher {
         if (CMSFeatureFlags.cmsFeatureFlags().isImagePublishingEnabled()) {
             imageFuture = publishImages(collection);
         }
-
         publishFilteredCollectionFiles(collection, collectionReader);
 
         if (CMSFeatureFlags.cmsFeatureFlags().isVerifyPublishEnabled()) {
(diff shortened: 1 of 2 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -124,8 +124,9 @@ public static boolean executePublish(Collection collection, CollectionReader col
         if (CMSFeatureFlags.cmsFeatureFlags().isImagePublishingEnabled()) {
             imageFuture = publishImages(collection);
         }
-
-        publishFilteredCollectionFiles(collection, collectionReader);
+        
+        Manifest manifest = Manifest.get(collection);
+        publishFilteredCollectionFiles(collection, collectionReader, manifest);
```
jessjenkins: Why are we getting the manifest outside the method and passing it in as a parameter? The method is already being passed the Collection so why no get the manifest in the method instead?

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -414,15 +415,15 @@ public static Map<String, String> createPublishingTransactions(Collection collec
      * @param collectionReader
      * @throws IOException
      */
-    public static void publishFilteredCollectionFiles(Collection collection, CollectionReader collectionReader)
+    public static void publishFilteredCollectionFiles(Collection collection, CollectionReader collectionReader, Manifest manifest)
             throws IOException {
         // We do not want to send versioned files. They have already been taken care of
```
jessjenkins: We might want to reword the comment to make it clear that it is filtering anything that is already copied via the manifest not just versioned files. OK it's all and only versioned files now but it won't be all versioned files after the next task we do.

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/publishing/Publisher.java
```diff
@@ -124,7 +124,7 @@ public static boolean executePublish(Collection collection, CollectionReader col
         if (CMSFeatureFlags.cmsFeatureFlags().isImagePublishingEnabled()) {
             imageFuture = publishImages(collection);
         }
-
+
```
jessjenkins: Tiny formatting issue due to whitespace addition

jessjenkins (approved):

PR #861, opened 2026-06-11
Feature flag collection write locking

lindenmckenzie:
### What

This PR feature flags write locking allowing us to continue using / developing it whilst removing the feature from production.

### How to review

Check looks ok - functions as normal. 

### Who can review

Not me.

### DIFF
--- a/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
+++ b/zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
@@ -574,7 +574,7 @@ public class Collection {
      * Releases the write lock on the collection if it is currently held.
      */
     private void unlockForWriting() {
-        if (this.hasWriteLock && this.shouldWriteLock) {
+        if (this.hasWriteLock) {
             info().data("path", this.path).log("releasing collection write lock");
             ReadWriteLock collectionLock = collectionLocks.get(this.path);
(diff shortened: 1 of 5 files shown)


zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
```diff
@@ -565,7 +556,36 @@ public Lock getWriteLock() {
...
+            }
+            this.hasWriteLock = true;
+        }
+    }
+
+    /**
+     * Releases the write lock on the collection if it is currently held.
+     */
+    private void unlockForWriting() {
+        if (this.hasWriteLock && this.shouldWriteLock) {
```
jessjenkins: Just a tiny comment, but is the second conditional even necessary as it can only ever be true when the first one is true. Belts and braces but just thought I'd mention.

jessjenkins (dismissed):

zebedee-cms/src/main/java/com/github/onsdigital/zebedee/model/Collection.java
lindenmckenzie: there is some edge condition where if the feature flag is switched on there can be a scenario where it doesn't have the write lock but is called to unlock. It's one of the things I need to get to the bottom of.

jessjenkins (approved):

PR #863, opened 2026-06-17
Add collection title to cache api payload

lindenmckenzie:
### What

Add collection name to cache API payload

### How to review

Check looks ok, tests pass.

To test it you'd need to run the `legacy-core-publishing` stack with this: https://github.com/ONSdigital/dp-legacy-cache-api/pull/59 and do a scheduled collection to see it get added to the cache records. 

### Who can review

Not me.


jessjenkins (approved):
🍰

